PR #3, opened 2020-11-19
Persist size of terminal across toggle

EdenEast:
Thanks for making this plugin. I could not find a simple way to create a popup/scratch terminal so I was about to write my own. 

This pr is adding persistent size to the terminal split. I use the popout terminal for things like compile checking my code. I find the the errors are longer then the terminal so I make it bigger to read. When I toggle it again it would change back to the size in the preference. 

Now before the split is hidden the size is saved to be later used when creating new open split.

persist

### DIFF
--- a/lua/toggleterm.lua
+++ b/lua/toggleterm.lua
@@ -17,6 +17,7 @@ local preferences = {
   size = 12,
   shade_filetypes = {},
   shade_terminals = true,
+  persist_size = true,
   direction = "horizontal"
 }
 
@@ -30,10 +31,17 @@ local persistent = {}
 ---   1. The size argument is a valid number > 0
 ---   2. There is persistent width/height information from prev open state
 ---   3. Default/base case perference size
+---
+--- If `preferences.persist_size = false` then option `2` in the
+--- list is skipped.
 --- @param size number
 local function get_size(size)
-  local persist_size = preferences.direction == "horizontal" and persistent['height'] or persistent['width']
-  return (size and size > 0) and size or persist_size and persist_size or preferences.size
+  if not preferences.persist_size then
+    return (size and size > 0) or preferences.size
+  end
+
+  local psize = preferences.direction == "horizontal" and persistent['height'] or persistent['width']
+  return (size and size > 0) and size or psize or preferences.size
 end
 
 local function create_term()
@@ -123,6 +131,7 @@ local function resize(size)
 
   vim.cmd(cmd .. " " .. size)
 end
+
 --- @param size number
 local function open_split(size)
   size = get_size(size)
--- a/lua/toggleterm.lua
+++ b/lua/toggleterm.lua
@@ -36,12 +36,13 @@ local persistent = {}
 --- list is skipped.
 --- @param size number
 local function get_size(size)
+  local valid_size = size ~= nil and size > 0
   if not preferences.persist_size then
-    return (size and size > 0) or preferences.size
+    return valid_size and size or preferences.size
   end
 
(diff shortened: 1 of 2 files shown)


lua/toggleterm.lua
```diff
@@ -24,6 +24,17 @@ local preferences = {
...
+local persistent = {}
+
+--- Get the size of the split. Order of priority is as follows:
+---   1. The size argument is a valid number > 0
+---   2. There is persistent width/height information from prev open state
+---   3. Default/base case perference size
+--- @param size number
+local function get_size(size)
+  local persist_size = preferences.direction == "horizontal" and persistent['height'] or persistent['width']
+  return (size and size > 0) and size or persist_size and persist_size or preferences.size
```
akinsho, line 36: I don't think the `persist_size and persist_size` duplication is necessary. Like `size or persist_size or preferences.size` haven't actually run it so don't know that I'm not missing something

akinsho (commented):
@EdenEast thanks for the PR, this feature looks good. I wonder if you could add a way to toggle this off though. I'm not sure what most people like or don't like. I think this is a good default but if someone wanted to be able to expand it temporarily but then have it be the original size on opening it again they might find this annoying.

EdenEast:
Added an option to set if you want persistent size to the preferences

lua/toggleterm.lua
```diff
@@ -32,8 +33,12 @@ local persistent = {}
 ---   3. Default/base case perference size
 --- @param size number
 local function get_size(size)
-  local persist_size = preferences.direction == "horizontal" and persistent['height'] or persistent['width']
-  return (size and size > 0) and size or persist_size and persist_size or preferences.size
+  if preferences.persist_size then
```
akinsho, line 36: Think this should be
```suggestion
  if not preferences.persist_size then
```

akinsho (commented):
Thanks for updating the docs as well completely forgot about that 🙌🏿 , just one last comment and this is good to go. Think the conditional for the persist size toggle is the wrong way round so setting it to `persist_size = true` will turn it off and vice versa

lua/toggleterm.lua
akinsho: actually also just realised it shouldn't directly return `preferences.size` it should return the `size` passed in to the function so someone can still call `exec(23)` and have the specific size they required open

EdenEast: Actually now that you mention that should it `return (size and size > 0) or preferences.size` so that it conforms to what was there before. Also yes will fix the not issue

lua/toggleterm.lua
```diff
@@ -17,13 +17,32 @@ local preferences = {
...
+---   1. The size argument is a valid number > 0
+---   2. There is persistent width/height information from prev open state
+---   3. Default/base case perference size
+---
+--- If `preferences.persist_size = false` then option `2` in the
+--- list is skipped.
+--- @param size number
+local function get_size(size)
+  if not preferences.persist_size then
+    return (size and size > 0) or preferences.size
```
akinsho, line 40: I think this will return the boolean if size is not nil rather than `size`
think it should look like the conditional below i.e.
```suggestion
    return (size and size > 0) and size or preferences.size
```

alternately to get past all the size checking you can save it to a variable and reuse it in both places i.e.
```lua
local valid_size = size ~= nil and size > 0

return valid_size and size or preferences.size
```

EdenEast: I like your solution better thanks will implement that

akinsho (approved):
Awesome LGTM 👍🏿

PR #5, opened 2021-01-13
`stopinsert' after hiding terminal

GenesisTMS:
When closing terminal from outside, it will leave you in mode that
you were in terminal.

example:
- open terminal - be in insert mode inside termnal
- move back to file - mode doesnt matter
- close terminal - and it will switch to insert mode

### DIFF
--- a/lua/toggleterm.lua
+++ b/lua/toggleterm.lua
@@ -406,6 +406,7 @@ function M.close(num)
     M.save_window_size()
 
     vim.cmd("hide")
+    vim.cmd("stopinsert!")
   else
     if num then
       vim.cmd(


GenesisTMS:
In general it would be nice if it would remember in which mode you were - and in which buffer :). - Working on it

Thanks for this plugin :)

akinsho:
@GenesisTMS thanks for the PR, LGTM 👍🏿 , gonna go ahead and merge this now if you're done.

Re. remembering your mode that seems kind of involved for not a huge amount of gain, that's fully a personal opinion though and I wouldn't say no to a PR if it was configurable and didn't involve adding too much complexity.

GenesisTMS:
Yes you can merge this.

as mode-remembering: i just thought that if you automatically map togglterm for normal as for insert mode - you should be left in these modes when you toggle term back.  
(stay in flow: `in the middle of writing <openterm> ... <closeterm> continue writing`)

don't know how much codechange it might be. Not going to try this for now - maybe later when it starts to bother me ;)

PR #6, opened 2021-01-13
Remember origin window to come back after close

GenesisTMS:
When you open a term it will save a current window. When closing it will
go back to that window if possible.
Also checks if you are in different window when closing term - so your
cursor is not jumping around.

Actual behaviour is that if you open term leaving it open and jump around windows - on close term your cursor jumps to nearest window.

### DIFF
--- a/lua/toggleterm.lua
+++ b/lua/toggleterm.lua
@@ -26,6 +26,7 @@ local preferences = {
 -----------------------------------------------------------
 local terminals = {}
 local persistent = {}
+local origin_win
 
 function M.save_window_size()
   -- Save the size of the split before it is hidden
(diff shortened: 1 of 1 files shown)


lua/toggleterm.lua
```diff
@@ -248,6 +256,9 @@ end
 --- @param size number
 local function toggle_nth_term(num, size)
   local term = find_term(num)
+
+  update_origin_win(term.window)
+
   if find_window(term.window) then
```
GenesisTMS, line 262: `update_origin_win` is here because `find_window` jumps to terminal window - so `origin_win` will be lost in `close` function.

akinsho:
Awesome just tried this out and it works great @GenesisTMS, hadn't realised how annoying the previous behaviour was till I tried it with you PR in vs the current state.

akinsho:
@GenesisTMS is this ready to merge?

GenesisTMS:
Eyop! :)
Thanks

Issue #8, opened 2021-02-15
shade_terminal isn't useful for light themes


urandom:
It produces a very low-contrast terminal that's very hard to see. A better approach might be to darken the terminal if the user uses a light theme instead.

akinsho:
@urandom thanks for raising this 👍🏿 can you share a screen shot +/- the color scheme name since I'm not sure what it looks like, light themes is a bit of a best effort since I never really use one other than the one that is part of `vim-one` once in a blue moon and that works fine with toggle term so that's what I based the light behaviour off of

urandom:
@akinsho 
Here's a link to a screenshot
https://imgur.com/cmyC84z

As you can see, the contrast is less than ideal. I assume the shade function reduces the lightness of the colors, which might be a wrong choice for light backgrounds, since the fonts are darker. A better result might be to increase the lightness instead.

I'm currently using the edge color scheme.

akinsho:
Hmm I'll have a play around to see what might work, I think increasing the lightness might amp up the brightness in an undesirable way. It might better instead to darken things even further for bright colorschemes so the contrast is greater

akinsho:
Going to close this now @urandom if the solution is workable for you

PR #9, opened 2021-02-15
Add ability to tweak degree of shading

akinsho:
This PR adds `shading_factor` to the available options so a user can tweak the amount of shading for the terminal buffer. It also darkens light backgrounds more by default.

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,7 @@ require"toggleterm".setup{
   open_mapping = [[<c-\>]],
   shade_filetypes = {},
   shade_terminals = true,
+  shading_factor = '<number>' -- the degree by which to darken to terminal colour, default: 1 for dark backgrounds, 3 for light
   start_in_insert = true,
   persist_size = true,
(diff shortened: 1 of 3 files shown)


akinsho:
@urandom this PR should hopefully address the light bg issue. They should now be darkened much more by default, and if this is not sufficient then there is a `shading_factor` which is used as a multiplier for the amount to shade the buffer it. Can you test it out and see if it works for you? I'm not really going to over correct for one colorscheme but hopefully this is fine

akinsho:
will fix #8

urandom:
by default, it seems to produce a dark background, but at least on my terminal, the command text is also darkish. Setting the shading factor to something small is workable though.

akinsho:
On my machine tbh the original seemed readable so don't know if that's an issue of how the terminal text is coloured 🤷🏿 anyway you can tweak the shading factor to your preference, I wont to delve into how the actual terminal text is coloured as I believe thats a can of worms

Issue #20, opened 2021-04-10
Ability to open a terminal in a separate buffer


niceloaf:
Is there a way you could open the terminal buffer in a separate buffer instead of a split?

akinsho:
@niceloaf do you mean a full window? since when the terminal is open it already creates a buffer. The buffer is just the file contents or in this case the shell. the window is what is what contains the content. In the case of toggle term, it creates a split window, so I'm guessing what youre asking is if it can be a full window not a split.

In which case basically no currently that isn't supported but someone could add this as new `direction` option maybe as both a tab and a window or just as a window to start. Shouldn't be too hard to do, not sure when I'll get to it myself, but happy to leave pointers if someone wants to look into this.

akinsho:
@niceloaf this plugin now offers a new `direction` which is `window` which will allow for creating a full window terminal. Fixed by #21

PR #21, opened 2021-04-12
         Refactor: Create terminal class API

akinsho:
This PR refactors a lot of the ad hoc logic from this plugins implementation and encapsulates it into a `Terminal` class, it also re-organises things so that creating new layouts such as opening a `window` or `tab` are easier.

This will also pave the way to having a floating window and a terminal api that users can use via
```lua
local Terminal = require('toggleterm.terminal').Terminal
local lazy_git_term  = Terminal:new{
  cmd = "lazygit",
  direction = "floating",
  dir = "<directory>",
}
lazy_git_term:toggle()
```

### TODO
- [x] toggling with a count should open that specific toggle term 
- [x] toggle term buffers should be resurrected post loading a session
- [x] validate `TermExec` works as expected
- [x] validate terminal size is persisted correctly
- [x] validate opening with existing layouts works
- [x] Add tests

### Moved to Follow up PR(s)
- Add floating window layout
- validate opening with dynamic layout is possible

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -1,0 +1,31 @@
+---
+name: Run tests
+on:
+  pull_request: ~
+  push:
+    branches:
+      - master
+
+jobs:
+  build:
+    name: Run tests
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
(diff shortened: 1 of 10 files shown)


akinsho:
fixes #20

PR #23, opened 2021-04-15
feature: add "float" display option

akinsho:
image

Following the refactor it is now easier to add more layouts. Next up is `float`, I envision the usage of this being for things like `lazygit` etc. Normally you want a toggle to popup and go away but sometimes you want it front and center like in a float. I don't usually like the overuse of floats but there are some places where things are best in a float

### TODO
- [x] Add options for customisation - before I get a zillion requests for it 😆 
	- [x] change highlights
	- [x] change borders - actually just expose internal option
	- [x] change size and location - again just expose what neovim provides 
- [x] Add tests to validate functionality


jan-xyz:
I am currently using https://github.com/voldikss/vim-floaterm but I was looking for a pure Lua implementation and found this! I am looking forward to this feature 😍

PR #27, opened 2021-04-22
Allow to chose custom shell

SamClercky:
Added an extra clause to allow another shell by simply setting vim.g.defaultShell = "myshell"

### DIFF
--- a/lua/toggleterm/terminal.lua
+++ b/lua/toggleterm/terminal.lua
@@ -179,7 +179,7 @@ end
 
 ---@private
 function Terminal:__spawn()
-  local cmd = self.cmd or vim.g.defaultShell or vim.o.shell
+  local cmd = self.cmd or config.get("shell")
   cmd = cmd .. ";#" .. term_ft .. "#" .. self.id
   self.job_id = fn.termopen(cmd, {
     detach = 1,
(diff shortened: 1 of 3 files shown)


lua/toggleterm/terminal.lua
```diff
@@ -179,7 +179,7 @@ end
 
 ---@private
 function Terminal:__spawn()
-  local cmd = self.cmd or vim.o.shell
+  local cmd = self.cmd or vim.g.defaultShell or vim.o.shell
```
akinsho, line 182: In able to this you'd need to
```lua
-- In config.lua add
config = {
  shell = vim.o.shell
}

-- In terminal.lua
local shell = config.get('shell')
local cmd = self.cmd or shell 
```

akinsho (changes requested):
@SamClercky thanks for the PR, I'm generally open to adding an option so users can configure this although I don't want any configuration to go through vim global variables though.

All configuration should go through the setup function. I've added a brief comment to give you an example of how that should work. You don't have to touch the setup function directly just add it to the configuration with the default value and fetch the key in terminal function

lua/toggleterm/terminal.lua
SamClercky: I have put the setting in the global config like you said :)

SamClercky:
In the last commit, I updated the README with the new setting

akinsho:
@SamClercky thanks this looks good 👍🏿 gonna test it once I get off work to confirm but otherwise happy to merge it

PR #33, opened 2021-05-03
Improve NeoVim [version] verification

kdav5758:
It simply checks if the user has NeoVim, if not, throws an error message, but if it does, it checks if it's nightly.

### DIFF
--- a/plugin/toggleterm.vim
+++ b/plugin/toggleterm.vim
@@ -1,8 +1,16 @@
 " Plugin file for terminal.vim
 " see: toggleterm/init.lua for more details
-if !has('nvim-0.5')
-  echoerr 'You need neovim nightly to run this plugin'
-  finish
+
+if has('nvim')
+	if !has('nvim-0.5')
+		echoerr 'You need Neovim Nightly to run this plugin'
+		finish
+	endif
+else
+    echohl Error
+    echom "Sorry this plugin only works with NeoVim ¯\_(ツ)_/¯"
+    echohl clear
+    finish
 endif
 
 "--------------------------------------------------------------------------------


plugin/toggleterm.vim
```diff
@@ -1,8 +1,16 @@
 " Plugin file for terminal.vim
 " see: toggleterm/init.lua for more details
-if !has('nvim-0.5')
-  echoerr 'You need neovim nightly to run this plugin'
-  finish
+
+if has('nvim')
```
akinsho, line 4: @kdav5758 Thanks for the PR, although I'm not sure why this extra nvim check is necessary? it already checks `nvim-0.5`? so I presume that would fail on everything that isn't nighly nvim including vim. Also tbh I think the README is pretty clear that this is an `nvim` plugin 🤷🏾 I'm not sure why someone would go this far having read the README (granted not everyone will)

kdav5758: > I'm not sure why someone would go this far having read the README

We all gotta hope for the best but prepare for the worst! As you said "granted, not everyone will". Trust me, while I was on Vim, I tried to install plugins written in Lua. If past me did this then it's likely that somebody else will too.

In the end it's up to you! personally I think it's nice to prepare for the worst-case scenario :)

akinsho: @kdav5758 I'm not against having some kind of block to check both neovim version *as well as vim* but the thing I'm not clear on from this change is whether `has('nvim-0.5')` returns true in `vim` since my guess when I originally added this was that it wouldn't. So I guess what I'm not sure of is what the additional check just for `nvim` alone adds?

I don't have vim installed just nvim but does running toggleterm in vim not trigger this warning or something?

kdav5758: `has('nvim-0.5')` will return `0` in vim but `1` in NeoVim. Here is an small demonstration of the whole thing:

test_has_nvim

akinsho: @kdav5758, thanks for the GIF 👍🏾 I'm still a little confused though, so the current way it's done in the plugin by just checking `has('nvim-0.5)` *will* catch both `vim` users *and* anyone using an nvim version less than `0.5` in which case why is any change needed?

It seems like the idea is to make the check more granular but I'm not sure why because atm if used in vim they will see an error saying they need neovim nightly, and if used in nvim-0.4 or less they will also see the error.

So what is different about this check? other than the error text? basically why have another conditional when one will catch both cases already?

kdav5758: The idea is simply to get a more "granular check", as you said, but at the same time, letting the user know what when wrong.

There are these three cases:
* If running on vim, throw an error.
* If running on a NeoVim  ver that's not nightly, throw an error.
* If running Nvim nightly, then you are good to go.

Indeed the plugin already catches all of these errors, but it doesn't specify to the user *what went wrong*, which is kind of the aim of this PR. :+1:

akinsho: @kdav5758 thanks for this PR I really appreciate contributions. In this case I don't think this is really necessary tbh. I think the original error message basically covers the same thing in slightly different language.

I think if someone reads `'You need Neovim Nightly to run this plugin'` or they read `"Sorry this plugin only works with NeoVim ¯\_(ツ)_/¯"` both of these things make it clear that the plugin requires neovim. I can't imagine someone reading *either* and not understanding that they need neovim. If they do somehow misunderstand the README also says this and there's only so far that I think it's worth going.

If they an nvim user it will clarify that they need the nightly, I don't think this change actually makes it clearer than the original message since (maybe I'm biased) but it already seemed pretty clear

kdav5758: Understandable, I'll close the PR then. :+1:

PR #36, opened 2021-05-09
feat: allow expanding vim wildcards

### LINKED ISSUE #35: Pass current file name to TermExec
As in #12, I'm trying to pass the current file name to `TermExec`.
However after the [syntax update](https://github.com/akinsho/nvim-toggleterm.lua#notices), `TermExec cmd='a'.expand('%').'b'` no longer works, it only considers `a` as the command.

akinsho:
This PR will allow using vim `wildcards` from `:h expand` inside the `cmd=<arg>` for example `cmd='echo %'` will be expanded to `cmd='echo /current/file/name.txt'`. Will fix #35

### DIFF
--- a/lua/toggleterm.lua
+++ b/lua/toggleterm.lua
@@ -49,6 +49,24 @@ function M.__apply_colors()
   end
 end
 
+---Expand wildcards similar to `:h expand`
+---all credit to @voldkiss for this vim regex wizadry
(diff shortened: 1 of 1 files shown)


akinsho:
@IDl0T can you try this PR again the bug you reported early should now be fixed. Can't guarantee there aren't other edge cases. Short of building a full on commandline arg parser that's bound to happen but this should handle most common cases

IDl0T:
It works fine for me!

PR #39, opened 2021-05-14
Check if terminal is open in `Terminal:shutdown`

l-kershaw:
Check if terminal is open before trying to close it in `Terminal:shutdown`.

Also allows `reset` to run without error even with closed terminals.


akinsho (approved):
@l-kershaw thanks for the PR 👍🏾 LGTM. FWIW since lua doesn't really have the notion of privacy a lot of the internal functionality other than those advertised in the README are subject to change (not planning anything but will do if needed, just so it doesn't explode for you in the future)

Issue #17, opened 2021-03-31
Question: Change direction dynamically?


hahuang65:
Is there a way for me to toggle a terminal, and then want to toggle another terminal, but in a different direction/orientation? It looks like right now, the direction is just hard-coded into the setup.

akinsho:
@hahuang65 no, you currently can't dynamically swap terminal orientation. I didn't actually add the vertical orientation that was another user. I don't really know what that'd be useful for tbh, anyway my approach with this plugin is if someone has an idea for a feature that they want to add that isn't too complex or different from the stated goals of the project I'm happy to take a look.

From my perspective this doesn't really fit into the the things that I feel need to be added so I don't think I'd add this myself.

hahuang65:
Thanks for the hard work so far! I'd love to see if I can make a PR but feel free to close this ticket if you want. I can always make a new PR ticket when I get around to it, if I get around to it.

akinsho:
@hahuang65 👍🏿 I'll keep this open for a little while in case you or someone else gets to it and close it eventually if no one does

zeertzjq:
Currently it seems different terminal can have different directions using `:ToggleTerm direction=`, however that only works for new terminals. The direction of existing terminals cannot be changed.

PR #40, opened 2021-05-16
Add option to give a function for the size

l-kershaw:
Allows users to give a function that returns the size wanted.
The option to simply give a number is kept.

Works for size of split and for width and height of float.

---

For example a user can give use the following in their setup:
```
⋮
size = function(term)
  if term.direction == 'horizontal' then
    return math.floor(vim.o.lines * 0.2)
  elseif term.direction == 'vertical' then
    return math.floor(vim.o.columns * 0.3)
  end
end,
⋮
```
To have horizontally split terminals take up 20% of the height, and vertically split terminals take up 30% of the width.

### DIFF
--- a/lua/toggleterm/ui.lua
+++ b/lua/toggleterm/ui.lua
@@ -1,6 +1,7 @@
 local M = {}
 
 local constants = require("toggleterm.constants")
+local utils = require("toggleterm.utils")
 
 local fn = vim.fn
 local fmt = string.format
@@ -184,16 +185,14 @@ local split_commands = {
 }
 
 --- @param size number/function
-local function resolve_size(size)
+function M._resolve_size(size,term)
   if type(size) == 'number' then
-    return function()
-      return size
-    end
-  end
-  if type(size) == 'function' then
     return size
   end
-  error(string.format('The input %s is not of type "number" or "function".',size))
+  if term and type(size) == 'function' then
+    return size(term)
+  end
+  utils.echomsg(string.format('The input %s is not of type "number" or "function".',size),"Error")
 end
 
 --- @param size number
@@ -202,8 +201,7 @@ function M.open_split(size, term)
   local has_open, win_ids = M.find_open_windows()
   local commands = split_commands[term.direction]
 
-  size = M.get_size(size)
-  size = resolve_size(size)(term)
+  size = M._resolve_size(M.get_size(size), term)
   if has_open then
     -- we need to be in the terminal window most recently opened
     -- in order to split to the right of it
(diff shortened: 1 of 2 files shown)


akinsho:
@l-kershaw thanks for the contribution. I hadn't done something like this already because I wasn't really sure what the use case would be.

A user can create terminal using the `Terminal` class and specify the `size` that way. They can also specify the size via `TermExec size=20`. What circumstances would require the size to be specified dynamically in the original configuration via a function?. I have a little logic in my config for something a bit similar but I just do the check outside of this plugin and pass the value to the config option. I guess this would help when something like #17 is (if ever) implemented.

Also will this interfere with those cases i.e. the exec and toggle commands etc.(seems like it wouldn't but I haven't check properly)

lua/toggleterm/ui.lua
```diff
@@ -183,13 +183,27 @@ local split_commands = {
...
+end
+
 --- @param size number
 --- @param term Terminal
 function M.open_split(size, term)
   local has_open, win_ids = M.find_open_windows()
   local commands = split_commands[term.direction]
 
   size = M.get_size(size)
+  size = resolve_size(size)(term)
```
akinsho, line 206: This is largely stylistic but I personally find the double invocation less readable than declaring another variable, granted it's slightly more wasteful but I think the syntax is less approachable.

lua/toggleterm/ui.lua
```diff
@@ -183,13 +183,27 @@ local split_commands = {
   },
 }
 
+--- @param size number/function
+local function resolve_size(size)
+  if type(size) == 'number' then
+    return function()
```
akinsho, line 189: It seems a bit pointless for this to return a function that will just return a number,
could this be something like
```lua
function resolve_size(size, term)
if type(size) == 'number' then
  return size
end
if term and type(size) == 'function' then
 return size(term)
end
```

lua/toggleterm/ui.lua
```diff
@@ -183,13 +183,27 @@ local split_commands = {
...
+      return size
+    end
+  end
+  if type(size) == 'function' then
+    return size
+  end
+  error(string.format('The input %s is not of type "number" or "function".',size))
```
akinsho, line 196: I try to avoid the use of `error` since it tends to print a pretty ugly message + stack trace to the message area I think it's nicer to print the message using `utils.echoerr` which spares the user details they likely don't care about

lua/toggleterm/ui.lua
akinsho: I think if you refactor this function to take two arguments then you could just wrap the result of `M.get_size` e.g.
```lua
size = resolve_size(M.get_size(size), term))
```

akinsho (changes requested):
@l-kershaw this generally looks good I've left some comments regarding the layout of resolve size i.e. I think it can be simplified. I also commented on the use case it's not too big a deal in either case since it's just an additional option but just want to clarify that it is needed.

Sorry a bit of a laundry list but I've also started to add tests in `toggleterm_spec.lua`, it'd be great to have a test case there to ensure that this functionality returns the right size and that the normal argument still works

l-kershaw:
Thanks for the feedback 🙂 

I have made the changes you suggested and have added in a few tests for the resolve function.
Let me know if this is sufficient on testing, or if there is anything else you want to be checked.

I don't think there will be any issues with the exec and toggle commands.
I haven't come across any, and I can't see any way for this change to cause problems for them.

---

In terms of use case, for me it is being able to setup the functions once at the beginning and know that using the `:ToggleTerm` command will still work as I want even when I change the size of my terminal window. Then I don't have to add a load of options on when using the commands.

akinsho (approved):
@l-kershaw thanks for making the changes and updating the tests, and clarifying your rationale (apologies if it seemed like the third degree 😆 )

akinsho:
Actually just pulled it down to give it one last check and it seems to cause errors when opening floating windows. The `size` variable is unused for floats and so is fine to be nil, but the function requires it to be a `number` or a `function`. I think if nothing is passed it should just return nothing

akinsho:
Thanks for the fix, works fine now 👍🏾

PR #41, opened 2021-05-17
Format cmd in OS compatible way

l-kershaw:
Currently when opening a new terminal and providing a command, this command is formatted using semicolons, ";".

- On Linux, this can be used as a separator between any list of commands (I think?).
- On Windows, some commands can be separated with semicolons, but others can't.
  Some commands will read the entire line, and not stop at semicolons (see [here](https://docs.microsoft.com/en-us/windows-hardware/drivers/debugger/----command-separator-#remarks)).
(In my use case, the "vifm" command line file manager reads the full line.)

Therefore it would be good to add an option to disable this formatting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,7 +205,6 @@ Each terminal can take the following arguments:
 ```lua
 Terminal:new {
   cmd = string -- command to execute when creating the terminal e.g. 'top'
-  no_format = boolean -- disables formatting of command input 
   direction = string -- the layout for the terminal, same as the main config options
   dir = string -- the directory for the terminal
   on_open = fun(t: Terminal) -- function to run when the terminal opens
--- a/README.md
+++ b/README.md
@@ -204,7 +204,8 @@ for showing terminal UIs like `lazygit`, `htop` etc.
 Each terminal can take the following arguments:
 ```lua
 Terminal:new {
-  cmd =  string -- command to execute when creating the terminal e.g. 'top'
+  cmd = string -- command to execute when creating the terminal e.g. 'top'
+  no_format = boolean -- disables formatting of command input 
   direction = string -- the layout for the terminal, same as the main config options
   dir = string -- the directory for the terminal
   on_open = fun(t: Terminal) -- function to run when the terminal opens
--- a/lua/toggleterm/terminal.lua
+++ b/lua/toggleterm/terminal.lua
@@ -9,6 +9,9 @@ local api = vim.api
 local fmt = string.format
 local fn = vim.fn
 
+local command_sep = has("win32") and "&" or ";"
+local comment_sep = has("win32") and "::" or "#"
+
 ---@type Terminal[]
 local terminals = {}
 
@@ -220,11 +223,14 @@ end
 ---@private
 function Terminal:__spawn()
   local cmd = self.cmd or config.get("shell")
-  if vim.fn.has("win32") then
-    cmd = fmt("%s&::%s::%d", cmd, term_ft, self.id)
-  else
-    cmd = fmt("%s;#%s#%d", cmd, term_ft, self.id)
-  end
+  cmd = table.concat({
+    cmd,
+    command_sep,
+    comment_sep,
+    term_ft,
+    comment_sep,
+    self.id
+  })
   self.job_id = fn.termopen(cmd, {
     detach = 1,
     cwd = _get_dir(self.dir),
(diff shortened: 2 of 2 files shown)


akinsho:
@l-kershaw this is actually a lot trickier than it seems. Each terminal is being appended with this as a way of finding and restoring terminal buffers after restarting a session something which was inspired by/stolen from neoterm. Since the name contains a toggleterm id that terminal on restarting nvim can just be picked back and updated as needed.

> When the terminal starts, the buffer contents are updated and the buffer is
named in the form of `term://{cwd}//{pid}:{cmd}`. This naming scheme is used
by |:mksession| to restore a terminal buffer (by restarting the {cmd})

I consider this must have functionality personally. On top of that it is also a mechanism by which I "resurrect" any terminal windows that have become disassociated somehow. It's also a fairly deep mechanism of the plugin so I definitely don't want to expose it as just another configuration option if there's another way round.

I would much rather that if the use of the separator is causing issues, maybe another separator character can be used safely or some other mechanism found to replace it.
neoterm seems to now use

https://github.com/kassio/neoterm/blob/1283da9f078669593f3828e4345b68b59ee9740f/autoload/neoterm/term/neovim.vim#L23

Maybe this is something that could be adapted 🤔 . I'd much rather a solution that still allowed finding lost toggleterms by name rather than more configuration.

l-kershaw:
Ah, okay. I'll have a look into it and see if I can find a different solution

l-kershaw:
I _think_ this fixes my issue without messing up any of the sessions stuff.
I checked everywhere a "#" or ";" was used in the codebase and added checks to replace them with "::" and "&" respectively on windows.

I don't really understand how the session stuff works in toggleterm, so I might have missed something.

lua/toggleterm/terminal.lua
```diff
@@ -298,7 +302,12 @@ end
 --- @return number
 function M.identify(name)
   name = name or api.nvim_buf_get_name(api.nvim_get_current_buf())
-  local parts = vim.split(name, "#")
+  local parts
+  if vim.fn.has('win32') then
+    parts = vim.split(name, "::")
```
akinsho, line 307: could this be determined once on startup since it's static otherwise e.g.
```lua
-- At the top of the file
local cmd_sep = has('win32') and "::" or "#"

--- then in the func
local parts = vim.split(name, cmd_sep)
```

lua/toggleterm/terminal.lua
```diff
@@ -220,7 +220,11 @@ end
 ---@private
 function Terminal:__spawn()
   local cmd = self.cmd or config.get("shell")
-  cmd = fmt("%s;#%s#%d", cmd, term_ft, self.id)
+  if vim.fn.has("win32") then
```
akinsho, line 223: Similar to the other comment if we determine once at the start what os someone is on then we can just use a separator variable although I think at this point `fmt` isn't the right way to do this and it should just be strings or table.concat to string 🤷🏾 just that '%s%s%s` etc is meh but not a big deal really fmt or not
```lua
--- at the top-ish
local is_windows = has('win32') -- no need to check this more than once
local cmd_sep = ... -- from other comment

--- re-use in separator here as well 🤷🏾 
cmd = ...
```

README.md
```diff
@@ -204,7 +204,8 @@ for showing terminal UIs like `lazygit`, `htop` etc.
 Each terminal can take the following arguments:
 ```lua
 Terminal:new {
-  cmd =  string -- command to execute when creating the terminal e.g. 'top'
+  cmd = string -- command to execute when creating the terminal e.g. 'top'
+  no_format = boolean -- disables formatting of command input
```
akinsho, line 208: Think this needs removing as well

akinsho (commented):
@l-kershaw thanks for finding an alternative solution 🚀 . I'll try this out later on tonight 👍🏾

akinsho:
@l-kershaw not sure if this ready or not but just gave it a quick try and the terminal closes immediately after opening on my end, not obvious error message though

l-kershaw:
@akinsho I don't get this issue on my end anymore.
I did at first (with the "close_on_exit" at the default value of true), which confused me for a while with testing commands like "echo test", but I think it's just down to that option.
It works as I would expect for things like "lazygit".

lua/toggleterm/terminal.lua
```diff
@@ -9,6 +9,9 @@ local api = vim.api
 local fmt = string.format
 local fn = vim.fn
 
+local command_sep = vim.fn.has("win32") and "&" or ";"
```
akinsho, line 12: @l-kershaw I believe this is the issue. `fn.has` returns a number so will always be truthy. I'm guessing you are on windows which is why it's working and I'm on linux where it's not

akinsho (changes requested):
@l-kershaw I think this is because the os check isn't working correctly and I'm getting windows separators on my machine which is breaking things

l-kershaw:
@akinsho This should _hopefully_ fix your issue

l-kershaw:
@akinsho This should hopefully fix your issue?

akinsho (approved):
Thanks for all your work on this. Tested it locally and it's working fine now 👍🏾

PR #44, opened 2021-05-26
fix: delete buffer on process exit

### LINKED ISSUE #43: Should delete buffer on process exit.
I made a custom floating buffer and if I close the shell by doing `exit`, the buffer is taken off focus but is still hidden. When I toggle the custom floating terminal again, it shows the last session with a `[Process exited 0]` at the end of terminal. The expected behaviour should be to delete buffer on process exit.

akinsho:
This should fix #43 by deleting the buffer when the process exits

### DIFF
--- a/lua/toggleterm/terminal.lua
+++ b/lua/toggleterm/terminal.lua
@@ -217,6 +217,9 @@ local function __handle_exit(term)
     end
     if config.get("close_on_exit") then
       term:close()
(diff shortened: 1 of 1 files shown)


akinsho:
@Chaitanyabsprip can you try this and see if it fixes this issue

Chaitanyabsprip:
This works absolutely fine. Thank you for fixing this and your efforts. Really enjoying the plugin and the customizability!

PR #50, opened 2021-06-06
fix: set `signcolumn` to `no` by default

weihanglo:
The signcolumn may be useless for a terminal window, so set it to no by default.

## How to test

1. open neovim
2. `:set signcolumn=yes`
3. `:ToggleTerm`

### DIFF
--- a/lua/toggleterm/ui.lua
+++ b/lua/toggleterm/ui.lua
@@ -45,6 +45,7 @@ function M.set_options(win, buf, term)
   end
   vim.bo[buf].buflisted = false
   vim.bo[buf].filetype = constants.term_ft
+  vim.wo[win].signcolumn = "no"
 
   local conf = require("toggleterm.config").get()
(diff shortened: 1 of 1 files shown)


akinsho:
@weihanglo I'm not too sure about adding this as a default. The signcolumn should be empty in a terminal buffer and tbh I think it provides a nice bit of padding so that the text in the terminal isn't flush with the border which doesn't look good IMO

image

Are signs being set on your machine somehow? just wondering why you would like it removed?
alternatively you could just create and autocommand in your `init.vim` `autocmd TermOpen * setlocal signcolumn=no`

weihanglo:
Your point is very reasonable to me. Thank for you quick reply and your alternative solution!

weihanglo:
BTW, on my machine I have a sign column with the same background as editing area, so it looks weird for the sign column occupying one column but do nothing.

image

PR #53, opened 2021-06-25
Use <Cmd> instead of :<c-u> in mappings

zeertzjq:
`<Cmd>` is more flexible and slightly shorter.


akinsho (commented):
@zeertzjq thanks for the PR, this change looks good. Will test it out a bit to make sure there aren't any obscure breakages, but otherwise 👍🏾

PR #56, opened 2021-07-08
fix: remove id from the list of available ids on terminal exit

toppair:
Ran into a problem where each newly opened terminal would have an incremental id 
assigned to it, even after all previous ones have been closed. These changes fixed 
things for me, so I'm making a pr instead of raising an issue, as requested.


akinsho:
Thanks for the PR @toppair, and for catching that. Will test it out locally when I get a chance, LGTM though 👍🏾

PR #58, opened 2021-07-10
feat: expand `self.dir` before opening a `Terminal`

l-kershaw:
Expand user-provided `dir` at the start of `Terminal:open`.

This then matches the setup for `require('toggleterm').exec`

### DIFF
--- a/lua/toggleterm.lua
+++ b/lua/toggleterm.lua
@@ -158,9 +158,6 @@ function M.exec(cmd, num, size, dir)
     size = { size, "number", true },
     dir = { dir, "string", true },
   })
-  if dir then
-    dir = fn.expand(dir)
-  end
   -- count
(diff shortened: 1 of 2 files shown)


akinsho:
@l-kershaw thanks for the PR 👍🏾 will review properly later. One thought I had after a cursory glance is whether or not there's a place that this could go to sort of centralise directory handling so there isn't ad hoc directory logic in different places. There is already `_get_dir` which handles expanding the special `dir = 'git_dir'` option, but I think that is only run for new terminals.

l-kershaw:
@akinsho I have rearranged things so that directory processing is routed through the `_get_dir` function.

Was this roughly what you had in mind?

akinsho (approved):
LGTM @l-kershaw thanks for making the changes 👍🏾

PR #69, opened 2021-07-23
Use <Cmd> instead of <C-\><C-N>: in terminal mode mapping

zeertzjq:
I didn't notice this in #53.
This change actually solves a bug: when mapping is `<C-t>` and it is mapped to `<C-\><C-n>:execute v:count . "ToggleTerm"<CR>`, pressing `<C-\><C-t>` types `<C-n>:execute v:count . "ToggleTerm"<CR>` in the terminal, and changing `<C-\><C-n>:` to `<Cmd>` solves this bug.


akinsho:
@zeertzjq thanks for raising the PR and for finding the bug 👍🏿

Issue #67, opened 2021-07-21
[Bug] horizontal term should close if it is the last window


qvieth:
E5108: Error executing lua ...ack/packer/opt/nvim-toggleterm.lua/lua/toggleterm/ui.lua:256: Vim:E444: Cannot close last window
This was thrown when I was trying to exit from a file buffer while a horizontal toggle term instance is opening
Expected behavior :wq on file buffer exit both file buffer and the horizontal toggle term.

qvieth:
file:///home/qvieth/Videos/Screencast%20from%2007-21-2021%2004:02:15%20PM.webm

qvieth:
https://user-images.githubusercontent.com/71810726/126462620-3f6f41ec-d95a-4df8-a3c1-8ba9aa140fc1.mp4

akinsho:
@qvieth this is partially a duplicate of #64, I've left it open though because a split window should technically not remain open if the other neighbouring window was closed. FWIW this is not how I envisioned people using the plugin or how I use it, till I or someone else gets a chance to fix this it is better if you *close the terminal* before closing the other buffer.

qvieth:
@akinsho Thank you, really appreciate your works on this plugin

akinsho:
The error should now no longer appear, but the automatic closing of the terminal if it is the last window doesn't work yet

qvieth:
@akinsho I just did an upgrade but that error still appeared as same as before

qvieth:
@akinsho i just did an upgrade but the error still appeared to be the same as before

akinsho:
Oh, sorry this isn't actually a duplicate of the error I referenced before, ignore that then the status of this is the same as before then.

akinsho:
@qvieth this should now be fixed by the most recent commit on master, it works by choosing another buffer using `bnext` if a user closes the neigbouring window

qvieth:
So cool! Thank you so much @akinsho, Toggleterm works flawlessly for me now!

PR #71, opened 2021-07-26
refactor: support dynamic window direction

### LINKED ISSUE #17: Question: Change direction dynamically?

### LINKED ISSUE #68: [Bug] Floating windows don't auto open on exec
Following up from https://github.com/akinsho/nvim-toggleterm.lua/issues/62#issuecomment-885187362

Summary: when I was using `require('toggleterm').exec(...)` to run something with [this configuration](https://github.com/ChristianChiarulli/LunarVim/blob/95aa1ea8c6b8dc31bae26fe1d433e83f57dd4201/lua/core/terminal.lua#L2), the floating terminal would not auto open 

@<i></i>akinsho said:
> Seems term exec doesn't open a float like it does for splits since it doesn't really seem to take the direction into consideration, maybe.

zeertzjq:
Closes #17, Closes #68


zeertzjq:
There is another problem: changes to terminal size are not always remembered. Maybe `size` can be assigned to a member variable of `Terminal` class to solve this problem?

akinsho:
@zeertzjq thanks for the PR, will have a look over the next couple of days. A few things to mention without having had a proper look first, I think go back is fine to make visible but should almost always default to true, at least for splits.

I think it's a massively valuable thing to be able to send things to the terminal without "moving" and I definitely wouldn't want to lose that because of float layouts, which tbh I only added because people pestered me 😅. 

It will also be a bit of a breaking change since people will have to go and tweak all their mappings to add go back. I think it should be default true everywhere but floats.

zeertzjq:
Maybe `go_back` can be ignored for floats and default to `true` for other cases.

zeertzjq:
Currently there are still some problems in changing terminal direction, related to `window` and `tab`.

zeertzjq:
It seems there is another existing bug (unrelated to this PR): after opening a vertical terminal, all horizontal terminals occupy the full screen.

akinsho:
@zeertzjq tab is actually not supported yet, it's mentioned but didn't exist. The issue with the sizing probably has to do with persisting the terminal size since that will no longer work with dynamic directions. It might need to be size persisted per terminal and direction instead

zeertzjq:
It indeed seems to be a problem of `persist_size`.

akinsho:
That functionality was added by another user a while ago when the scope of this plugin was simpler and we have envisioned all the layouts this plugin now has. It's definitely due a refactor

zeertzjq:
I guess I'll leave size handling unchanged in this PR.

akinsho:
> I guess I'll leave size handling unchanged in this PR.
I think it depends on how you decide to scope this PR, if it is primarily around fixing issues like being able to use term exec with floats, then I'm fine with leaving size handling out.

I don't think though that if this PR advertises dynamic positioning in the README that it should be broken, since atm without the size handling changes anyone who tries to change the terminal's direction will hit bugs which will inevitably lead to a bunch of issues. I'd rather that if the size stuff still doesn't work right, then this PR shouldn't mention that functionality as it isn't ready for users yet then.

zeertzjq:
> > I guess I'll leave size handling unchanged in this PR.
> 
> I think it depends on how you decide to scope this PR, if it is primarily around fixing issues like being able to use term exec with floats, then I'm fine with leaving size handling out.
> 
> I don't think though that if this PR advertises dynamic positioning in the README that it should be broken, since atm without the size handling changes anyone who tries to change the terminal's direction will hit bugs which will inevitably lead to a bunch of issues. I'd rather that if the size stuff still doesn't work right, then this PR shouldn't mention that functionality as it isn't ready for users yet then.

I'm sorry I wasn't thinking that much. I just thought this PR's title is long enough, and if I add more commits I'll have to make the title even longer to summarize the content of the PR.

akinsho:
I don't really mind about what the PR title is tbh can always just add any extra information in the commit description.

zeertzjq:
> There is another problem: changes to terminal size are not always remembered. Maybe `size` can be assigned to a member variable of `Terminal` class (and probably set to `nil` (to use the default value) if `direction` is changed) to solve this problem?

This plugin current saves window size on `CursorHold`, and maybe that's why changes to terminal size are not always remembered.

zeertzjq:
I've made `persist_size` work with multiple directions, but not for splits. For splitting to work correctly with multiple directions I guess `find_open_windows` needs to be rewritten.

README.md
```diff
@@ -138,18 +138,24 @@ This plugin provides 2 commands
...
+If `dir` is specified on creation toggle term will open at the specified directory.
+
 The directory can also be specified as `git_dir` which toggleterm will then
 use to try and derive the git repo directory.
 *NOTE*: This currently will not work for work tree's or other more complex setups
 
-If specified on creation toggle term will open at the specified directory at the
-specified height.
+If `size` is specified on creation toggle term will open at the specified height/width
+(doesn't apply to direction=window/tab/float). If the terminal has been created before
```
akinsho, line 154: Can you remove the mention of `tab` in the possible directions here. It's fully unsupported atm, just something I toyed with implementing but don't want someone to see this and raise an issue about why it doesn't work since it's not expected to.

lua/toggleterm.lua
```diff
@@ -60,14 +60,14 @@ end
 --Create a new terminal or close beginning from the last opened
 ---@param _ number
 ---@param size number
----@param directory string
+---@param dir string
 ---@param direction string
-local function smart_toggle(_, size, directory, direction)
+local function smart_toggle(_, size, dir, direction)
```
akinsho, line 65: This is a fairly minor point but why rename this to `dir` rather than directory, tbh the only reason I've used dir anywhere was to make things a little more ergonomic so users don't have to type `directory` all the time. I don't do the same with direction because of the clash and because frankly the direction flag was a bit of an afterthought

lua/toggleterm/commandline.lua
```diff
@@ -41,6 +41,8 @@ function M.parse(args)
         local key, value = arg[1], arg[2]
         if key == "size" then
           value = tonumber(value)
+        elseif key == "go_back" then
+          value = tonumber(value) ~= 0
```
akinsho, line 45: Why use `0` and `1` here rather than the string true and false, seems clearer to just have an actual boolean even though that would be converted to a string and would need to be changed back to a boolean 🤷🏿‍♂️ not a big deal though

lua/toggleterm/terminal.lua
```diff
@@ -110,7 +110,7 @@ local function setup_buffer_autocommands(term)
     table.insert(commands, {
       "CursorHold",
       fmt("<buffer=%d>", term.bufnr),
-      "lua require'toggleterm.ui'.save_window_size()",
+      fmt("lua require'toggleterm.ui'.save_window_size(0, '%s')", term.direction),
```
akinsho, line 113: nit: maybe in this case it's better to pass nil to the save window size's first arg and then in the function if the window is nil then use `nvim_get_current_win` so that it uses the window id rather than depending on the `0` is current behaviour

lua/toggleterm/terminal.lua
```diff
@@ -298,15 +305,21 @@ local function opener(size, term)
     ui.open_tab(term)
   elseif direction == "float" then
     ui.open_float(term)
+  else
+    error("Invalid direction")
```
akinsho, line 309: I prefer to use the util functions rather than `error` since error prints a pretty ugly stack trace which I find fairly distracting. Actually long term I want to migrate all errors and messages to use
```lua
vim.notify(msg, vim.log.levels.ERROR)
```

lua/toggleterm/ui.lua
```diff
@@ -10,10 +10,15 @@ local api = vim.api
...
-function M.save_window_size()
-  -- Save the size of the split before it is hidden
-  persistent.width = vim.fn.winwidth(0)
-  persistent.height = vim.fn.winheight(0)
+--- Save the size of a split window before it is hidden
+--- @param window number
+--- @param direction string
+function M.save_window_size(window, direction)
+  if direction == "horizontal" then
+    persistent.horizontal = vim.fn.winheight(window)
```
akinsho, line 18: This I think can actually be migrated to use api functions for a probably negligible speed increase e.g. `vim.api.nvim_win_get_{height|width}`

README.md
```diff
@@ -93,7 +93,7 @@ require("toggleterm").setup{
   start_in_insert = true,
   insert_mappings = true, -- whether or not the open mapping applies in insert mode
   persist_size = true,
-  direction = 'vertical' | 'horizontal' | 'window' | 'float',
+  direction = 'vertical' | 'horizontal' | 'window' | 'tab' | 'float',
```
akinsho, line 96: `Tab` isn't supported yet, so I don't think this line is needed.

akinsho (commented):
This generally looks good @zeertzjq just left some minor comments 👍🏿

lua/toggleterm.lua
akinsho: I guess maybe it's for consistency in which case that's fine

akinsho:
> I've made specifying size= work with multiple directions. persist_size also works in this case,

Sounds great thanks 👍🏿 
> but not when the terminals are manually resized (CursorHold is not very reliable).

Tbh this was a user submitted change that has been a bit of a pain to maintain as this plugin has changed I don't think it needs to be optimised too much, but I guess it could use a `WinResized` autocommand instead of `CursorMoved` 🤷🏿‍♂️. I really am personally not fussed as long as it doesn't break things.


> BTW, splitting currently doesn't work if there is an existing window with a different direction. I guess find_open_windows() has to be rewritten to fix that.

Yeah this is one of the reasons I've been wary of trying to do dynamic window stuff, basically I really don't want to go too far down the road of also building a split manager type thing that's checking the orientation of stuff in too much detail etc. I personally really wish people would just pick a direction and use that since that's how I use this plugin except for the occasional float.

I think as long as that caveat is mentioned in the README that's fine. If someone else wants to contribute a PR to fix that then they are welcome to 🤷🏿‍♂️

akinsho:
I simplified the PR title btw, can add specific details on the changes in the commit message for the PR

lua/toggleterm/terminal.lua
zeertzjq: The problem is `Terminal:open()` cannot proceed if the direction is invalid.

zeertzjq: The problem is `Terminal:open` cannot proceed if the direction is invalid.

lua/toggleterm/terminal.lua
zeertzjq: I'll switch the order of the two parameters then.

lua/toggleterm/terminal.lua
akinsho: @zeertzjq hmm in which case I guess it's fine for now I'll look into a different solution (like returning an `ok` value which is a success boolean and handling things in the caller e.g. an early return) at another time since I really don't like the way it presents errors, but it's not a blocker.

lua/toggleterm/ui.lua
```diff
@@ -10,10 +10,16 @@ local api = vim.api
...
 
-function M.save_window_size()
-  -- Save the size of the split before it is hidden
-  persistent.width = vim.fn.winwidth(0)
-  persistent.height = vim.fn.winheight(0)
+--- Save the size of a split window before it is hidden
+--- @param window number
+--- @param direction string
+function M.save_window_size(direction, window)
+  window = window or nvim_get_current_win()
```
akinsho, line 17: this looks like its missing a `vim.api.nvim_get_current_win`, so will error if the window passed in is nil

zeertzjq:
> > I've made specifying size= work with multiple directions. persist_size also works in this case,
> 
> Sounds great thanks 👍🏿
> 
> > but not when the terminals are manually resized (CursorHold is not very reliable).
> 
> Tbh this was a user submitted change that has been a bit of a pain to maintain as this plugin has changed I don't think it needs to be optimised too much, but I guess it could use a `WinResized` autocommand instead of `CursorMoved` 🤷🏿‍♂️. I really am personally not fussed as long as it doesn't break things.
> 
> > BTW, splitting currently doesn't work if there is an existing window with a different direction. I guess find_open_windows() has to be rewritten to fix that.
> 
> Yeah this is one of the reasons I've been wary of trying to do dynamic window stuff, basically I really don't want to go too far down the road of also building a split manager type thing that's checking the orientation of stuff in too much detail etc. I personally really wish people would just pick a direction and use that since that's how I use this plugin except for the occasional float.
> 
> I think as long as that caveat is mentioned in the README that's fine. If someone else wants to contribute a PR to fix that then they are welcome to 🤷🏿‍♂️

`WinResized` autocommand doesn't exist yet. I'll try to size `BufLeave`.

zeertzjq:
I think I've made `persist_size` work for manually resized terminals. However, there seems to be some problem with the rendering of the terminal being split.

zeertzjq:
`close_last_window()` seems to always cause an error.

akinsho:
@zeertzjq the point of `close_last_window` is to solve issues like #67 basically it's similar to the NERDTree autocommand that used to be added to people's dotfiles i.e. if a split term is open and the neigbouring window with the users file is closed then rather than expanding to be the only full size window the terminal should also close. Ideally, it navigates to another previous buffer

README.md
```diff
@@ -138,22 +138,31 @@ This plugin provides 2 commands
...
+
+If `direction` is specified and the command opens a terminal,
+the terminal will be changed to the specified direction.
 
 _NOTE_: If the terminal has already been opened at a particular directory it will
 remain in that directory.
 
+_NOTE_: `size` and `direction` are ignored if the command closes a terminal.
+
+_NOTE_: Support for having terminals with different directions open at the same time is currently buggy.
```
akinsho, line 164: I think having 3 separate note sections is a bit strange. I think these can all be consolidated into on heading that just explains things to note, like `Caveats`. Also, I think the wording that something is buggy doesn't sound very good. I'd rather just say it's not supported currently.

README.md
```diff
@@ -168,6 +177,13 @@ note that the `cmd` argument **must be quoted**.
 
 The `cmd` and `dir` arguments can also expand the same special keywords as `:h expand` e.g.
 `TermExec cmd="echo %"` will be expanded to `TermExec cmd="echo /file/example"`
+
+The `size` and `direction` arguments are like the `size` and `direction` arguments of `ToggleTerm`.
+
+By default focus is returned to the original window after executing the command
+(except for floating terminals).
+Use argument `go_back=0` or `go_back=false` to disable this behaviour.
```
akinsho, line 185: Sorry I should have been clearer, I think go back should be one type of value or the other. I think it taking both a number and a false or true string is too many permutations. I just meant it should be true or false, no numbers.

lua/toggleterm.lua
```diff
@@ -110,11 +110,20 @@ end
 function M.handle_term_enter()
   local _, term = terms.identify()
   if term then
-    term:resize()
     close_last_window(term)
   end
 end
 
+function M.handle_term_leave()
```
akinsho, line 117: What is this function for? Presumably to close floats, but I can't see why this PR needs it

lua/toggleterm.lua
```diff
@@ -110,11 +110,20 @@ end
 function M.handle_term_enter()
   local _, term = terms.identify()
   if term then
-    term:resize()
```
akinsho, line 113: Why remove this? This is there to prevent nvim forcefully resizing the window because of settings like `equalalways` etc.

lua/toggleterm/terminal.lua
```diff
@@ -106,13 +106,7 @@ local function setup_buffer_autocommands(term)
       "startinsert!",
     })
   end
-  if conf.persist_size and term:is_split() then
```
akinsho, line 109: This seems to me like persist size is no longer being respected and no longer specific to just terminals

akinsho (commented):
@zeertzjq there's quite a few new changes that, I'm not sure if I understand the rationale for. I wonder if it might be worth scoping down this PR maybe to a few more incremental changes 🤔 I know that in general there's not that much changed, but I worry a lot of seemingly core bits of logic have now been changed and this could result in regressions.

lua/toggleterm/terminal.lua
zeertzjq: I forgot to check the value of `persist_size` where window sizes are saved.

lua/toggleterm.lua
zeertzjq: It is mainly used to save window sizes before they are closed.

akinsho: We should only bother saving sizes if a user wants us to persist size though.

zeertzjq: I move closing floats here so that if a terminal is changed from a float to another direction, it is no longer automatically closed. A buffer `autocmd` was used before.

README.md
zeertzjq: But `go_back=0` is easier to type.

akinsho:
Just tried it out and it seems to be working pretty nicely 👍🏿 haven't seen any issues yet. I'll try it out locally for a bit and look more closely at the changes but seems pretty good so far.

lua/toggleterm.lua
zeertzjq: I removed this so that manual resizing while the window is not focused can be remembered. Is there a way to distinguish between automatic resizing and manual resizing?

README.md
akinsho: @zeertzjq tbh I don't imagine people will be manually typing out `go_back` in the cli often and are more likely to map it so something. Tbh I hate the vim convention of 0 as falsy and 1 or greater as true. I'd much rather this was just clear than necessarily just easy to type.

Another potential approach would be having a flag value something like `--focus-term` or something so just the presence of the flag is taken as truthy.

lua/toggleterm.lua
akinsho: That makes sense 👍🏿

lua/toggleterm.lua
zeertzjq: Isn't that already done in `open_split()` even before this PR?

README.md
zeertzjq: Maybe it can be changed to just a string `focus` instead.

lua/toggleterm.lua
akinsho: @zeertzjq this is here also to catch them case where someone is using vim sessions and the terminal gets "resurrected" vim might not open it to correct size on vim enter

README.md
akinsho: Regarding brevity, one day I hope to add commandline completion for the plugin so tbh I don't think the length of the keys will be that big a deal once a user can tab complete but definitely don't want this to have 2 different options.

I don't think a naked string argument is a good idea either I want the option to the commandline arguments to be `key=value` so parsing can be consistent. I really don't think the 3-4 characters saved is worth it here for an option that I feel confident that the vast majority of users will not be typing since it wasn't even available beforehand. I don't mind it being exposed but not at the cost of using numbers as booleans.

lua/toggleterm.lua
zeertzjq: I guess it is safe not to call this if `persistent[term.direction] ~= nil` then.

akinsho: Yeah I guess that works since if there is a persistent size already then it is not a newly resurrected terminal.

README.md
akinsho: You know what, it's fine, you can just change it to `go_back=0|1` I personally hate this pattern so might still change it in the future but it's not that big a deal.

zeertzjq:
Currently it seems unnecessary to call `save_window_size()` in `handle_term_leave()`. I guess sizes need to be saved more frequently to make resurrecting work properly.

zeertzjq:
I've made it that `Terminal:resize()` saves sizes while `resize_split()` does not.

lua/toggleterm.lua
```diff
@@ -6,6 +6,8 @@ local constants = require("toggleterm.constants")
 local colors = require("toggleterm.colors")
 
 local terms = require("toggleterm.terminal")
+local ui = require("toggleterm.ui")
```
akinsho, line 9: @zeertzjq in order to minimize startup time lua plugins should defer requiring as much as possible till the point where something is needed. I've purposely required things like config inside the functions so that as much as possible requires don't happen at the top of the file since that means just by a user requiring the setup function they basically start a nested chain of requires.

akinsho: Also tbh it just adds more visual noise to the PR to have changes that aren't related explicitly to what this PR is trying to achieve so it will be harder for me to properly review.

zeertzjq:
Maybe functions called from autocommands can also be moved out of `toggleterm.lua` then?

akinsho:
@zeertzjq I'm not sure which ones you mean but that sounds like an unrelated change and there are already a lot of changes here. I think those should stay there primarily because it is easier to export those functions so I can do `command = 'lua require'toggleterm'.my_autocommand_func()` rather than digging around in random modules for those.

In either case I think it's out of scope for this PR to move those functions and I'm planning on changing how the autocommand functionality works to accept lua functions directly.

lua/toggleterm.lua
```diff
@@ -110,11 +110,24 @@ end
...
   end
 end
 
+function M.handle_term_leave()
+  local _, term = terms.identify()
+  if term then
+    if term:is_float() then
+      term:close()
+    end
+  end
```
akinsho, line 128: this can be combined I think to
```suggestion
if term and term:is_float() then
  term.close()
end
```

lua/toggleterm/commandline.lua
```diff
@@ -41,6 +41,8 @@ function M.parse(args)
         local key, value = arg[1], arg[2]
         if key == "size" then
           value = tonumber(value)
+        elseif key == "go_back" then
+          value = value ~= "false" and value ~= "0"
```
akinsho, line 45: You can just stick with `0|1` as I mentioned above I might change it but it's a minor point in the grand scheme of things

akinsho:
@zeertzjq thanks for persisting with this 👍🏿 It generally looks good. The only things I still think need changing are:

- [ ] the multiple notes in the README
- [ ] `go_back` can just be 1 or 0 for now 

I will test it out later on today and check that no use cases have broken and if it's all good when will merge 👍🏿

akinsho:
@zeertzjq just tried this out quickly locally and tried moving from `vertical` to `horizontal` to `float` and then when I try to close the float using my toggle mapping I get
```
Error detected while processing WinLeave Autocommands for "term://*toggleterm#*":
E5108: Error executing lua ...rt/local-nvim-toggleterm.lua/lua/toggleterm/terminal.lua:188: attempt to index local 'self' (a nil value)
```

zeertzjq:
Maybe it's cause by a typo `term.close()`, I changed it to `term:close()`

zeertzjq:
> @zeertzjq just tried this out quickly locally and tried moving from `vertical` to `horizontal` to `float` and then when I try to close the float using my toggle mapping I get
> 
> ```
> Error detected while processing WinLeave Autocommands for "term://*toggleterm#*":
> E5108: Error executing lua ...rt/local-nvim-toggleterm.lua/lua/toggleterm/terminal.lua:188: attempt to index local 'self' (a nil value)
> ```

Maybe it's cause by a typo `term.close()` which should be `term:close()`.

zeertzjq:
Maybe it's caused by a typo `term.close()` which should be `term:close`. Is it fixed now?

akinsho:
@zeertzjq yep, that was the issue `:` method syntax implicitly passes `self` as the first argument whereas `.` doesn't

lua/toggleterm/ui.lua
```diff
@@ -318,7 +329,8 @@ end
 ---@param size number
 function M.resize_split(term, size)
   if term:is_split() then
-    size = M._resolve_size(size or M.get_size(), term)
+    local persist_size = require("toggleterm.config").get("persist_size")
```
akinsho, line 332: This value seems to be unused, within this function

lua/toggleterm/terminal.lua
```diff
@@ -175,6 +169,9 @@ end
 function Terminal:resize(size)
   if self:is_split() then
     ui.resize_split(self, size)
+    if config.get("persist_size") then
+      ui.persistent[self.direction] = size
```
akinsho, line 173: Looking through this locally, I wonder if all the persistence logic can't be encapsulated inside the `ui` module, so no other parts of toggleterm need to really think about persisting things, e.g. can this logic be moved into `resize_split` itself? and other logic that directly checks or manipulates the persistence table.

In this case there are a few places where the size is set, but it depends on exposing persistent and adding values to it which exposes the implementation detail of the table etc. I think this would be better as a setter that is just a wrapper around adding to the table but means in the future it can be changed without needing to touch this file

e.g.
```lua
M.persist_size(self)

function M.persist_size(term, size)
 -- this way persistent can be made private and the implementation changed in the future
 -- without affecting other areas of code
 persistent[term.direction] = size 
end
```

That might be a bit garbled let me know if what I'm saying is clear or not, I can always change it myself after, but I just think having persistence related logic in different modules will make changing its behaviour in the future or refactoring it later harder.

lua/toggleterm.lua
```diff
@@ -110,11 +110,22 @@ end
 function M.handle_term_enter()
   local _, term = terms.identify()
   if term then
-    term:resize()
+    local ui = require("toggleterm.ui")
+    local persist_size = require("toggleterm.config").get("persist_size")
+    if term:is_split() and not (persist_size and ui.persistent[term.direction]) then
```
akinsho, line 115: Here is another example where I wonder if the specifics of `persistent` can't be encapsulated inside of resize_split i.e. can that function do the checking if it should bother persisting sizes or not then this module doesn't have to deal with directly checking the persistence table

lua/toggleterm/terminal.lua
zeertzjq: That sounds like a good idea.

lua/toggleterm.lua
```diff
@@ -110,11 +110,21 @@ end
 function M.handle_term_enter()
   local _, term = terms.identify()
   if term then
-    term:resize()
+    if term:is_split() then
+      -- On session resurrect, neovim may incorrectly resize splits after TermOpen
+      require("toggleterm.ui").resize_split(term, nil, true)
```
akinsho, line 115: maybe `resize_split` should take a table so its arguments are clear e.g.
```lua
require("toggleterm.ui").resize_split { term = term, from_session = true, size = 20} 
```

lua/toggleterm/terminal.lua
```diff
@@ -175,6 +169,9 @@ end
 function Terminal:resize(size)
   if self:is_split() then
     ui.resize_split(self, size)
+    if config.get("persist_size") then
+      ui.save_direction_size(size, self.direction)
```
akinsho, line 173: This still leaks persistence implementation details and also doesn't look necessary anymore since resize split now already calls `ui.save_direction_size` right? the `resurrect_only` argument can just be renamed to `save_size` or something using a named args table and set to `true` for session restore as well as terminal resizing

lua/toggleterm/ui.lua
```diff
@@ -7,13 +7,26 @@ local fn = vim.fn
...
 local api = vim.api
 
-local persistent = {}
 local origin_window
 
-function M.save_window_size()
-  -- Save the size of the split before it is hidden
-  persistent.width = vim.fn.winwidth(0)
-  persistent.height = vim.fn.winheight(0)
+persistent = {}
```
akinsho, line 12: Without the `local` keyword this variable is now global 😅 , it shouldn't need to be if it is now only referenced in this module

akinsho (commented):
Thanks for the tweaks, this now really nicely hides the persistence implementation details. Just one point where the details still leak through and can be simplified as well as some style points around an accidental global variable and using named args for `resize_split`

lua/toggleterm/ui.lua
zeertzjq: Oops, I forgot to add `local`.

lua/toggleterm/terminal.lua
zeertzjq: But it is necessary to check whether `persistent[term.direction]` is `nil` to determine whether it is really session restore.

akinsho: I'm not sure I follow, there isn't a place inside this function where that check is being done though, maybe what I was getting at isn't clear.

I just mean calling save direction inside of the terminal means that it still references stuff to do with persistence but when I read through the implementation of `resize_split` it can already call `save_direction_size` if the `resurrect_only` argument is true so since it is already being called in resize split then why call it again here rather than just call resize split with the third arg as true.

But in this case it's confusing to call it `resurrect_only` it should just be called `save_size` or something. Does that make sense?

zeertzjq: I guess I need to change `Terminal:__resurrect()` to eliminate the need for `resurrect_only`.

akinsho: @zeertzjq checking the config value is fine but I think maybe what I mean isn't clear. I've explained a bit more below. It could be that I'm missing something though

lua/toggleterm/ui.lua
```diff
@@ -316,11 +333,18 @@ end
 ---Resize a split window
 ---@param term Terminal
 ---@param size number
-function M.resize_split(term, size)
-  if term:is_split() then
-    size = M._resolve_size(size or M.get_size(), term)
-    vim.cmd(split_commands[term.direction].resize .. " " .. size)
+---@param resurrect_only boolean
+function M.resize_split(term, size, resurrect_only)
```
akinsho, line 337: @zeertzjq so what I'm suggesting is
```lua
function M.resize_split(opts)
 local term, size, save_size = opts.term, opts.size, opts.save_size
 --- ... other lines in between then
 -- sizes have not been saved on session resurrect
 if save_size and persist_size and persistent[term.direction] then
  return
 end
 size = M._resolve_size(M.get_size(size, term.direction), term)
 if save_size then
   M.save_direction_size(size, term.direction)
 end
end
```

So the two places where you would set `save_size` are when resurrecting and when resizing split and so you don't need to call `save_direction_size` outside this module? Will that work or am I missing something

zeertzjq:
I've made it work again. Too tired to explain more for now.

akinsho:
@zeertzjq thanks, it's great as is 👍🏿 can always look into that later if needed. Will merge after a few tests later on today

akinsho:
Thanks again for all your hard work on this @zeertzjq 🚀

will:
Thanks @zeertzjq and @akinsho !!

PR #72, opened 2021-07-26
feat: add close_on_exit field to the Terminal class

grvxs:
Add a `close_on_exit` field to the Terminal class.
If not specified it uses the global `close_on_exit` specified in the config.

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,6 +219,7 @@ Terminal:new {
   cmd = string -- command to execute when creating the terminal e.g. 'top'
   direction = string -- the layout for the terminal, same as the main config options
   dir = string -- the directory for the terminal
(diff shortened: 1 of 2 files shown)


akinsho:
@grvxs thanks for the PR, can you explain a little about why you've raised it? Was there a particular problem you were trying to solve?

grvxs:
I run my code in neovim. For example I'm printing some text. In this situation close_on_exit cannot be true because I want to see the output. But there are situations where I want to exit the terminal when the process exits like lazygit. So instead of having one global option to control all the terminals, we could have a field in the terminal class to control closing of terminal windows.

akinsho:
That make sense @grvxs thanks for clarifying 👍🏿

PR #87, opened 2021-08-18
feat: support sending commands without opening window

### LINKED ISSUE #55: send commands to terminal without toggle
Is there a way to send commands to terminal without opening ?
Thank you

hexium310:
This PR resolves #55.

Even if a terminal window isn't open, sending commands with `chansend()` is available. The `open` parameter for whether open a terminal window or not is added to `Terminal.exec()`. This is used as follows:
```vim
:TermExec cmd='echo "hello world"' open=0
```


akinsho (approved):
@hexium310 this LGTM 🚀 
thanks for adding it 👍🏿

PR #88, opened 2021-08-22
refactor: remove bang from startinsert command

JA-Bar:
When using a floating terminal, opening and closing under certain scenarios results in a shift in the buffer's contents to the left.
To reproduce open a terminal with "float" direction, run `lazygit` and toggle the term a couple of times. The terminal's content will shift.

The problem may be an issue with using `startinsert!` as an autocmd, refactoring it to be executed once the terminal is open as a regular conditional fixes the issue.

If the root cause of the problem is related to other things let me know, but I believe this solution shouldn't sacrifice performance or code consistency.

### DIFF
--- a/lua/toggleterm/terminal.lua
+++ b/lua/toggleterm/terminal.lua
@@ -89,6 +89,7 @@ end
 ---Terminal buffer autocommands
 ---@param term Terminal
 local function setup_buffer_autocommands(term)
+  local conf = config.get()
   local commands = {
     {
       "TermClose",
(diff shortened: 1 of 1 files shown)


lua/toggleterm/terminal.lua
```diff
@@ -98,15 +97,6 @@ local function setup_buffer_autocommands(term)
     },
   }
 
-  if conf.start_in_insert then
-    vim.cmd("startinsert!")
-    table.insert(commands, {
-      "BufEnter",
-      fmt("<buffer=%d>", term.bufnr),
-      "startinsert!",
```
akinsho, line 106: @JA-Bar is the issue resolved by removing the `!` from the command. It actually causes nvim to behave as if `A` was pressed, which isn't what I intended.

akinsho: Rather than adding this logic to the `opener`

JA-Bar: Oh you are correct! I didn't think about it being unintended, should be correct now.

akinsho:
Thanks for the fix @JA-Bar 🚀

PR #96, opened 2021-09-18
feat: support both cmd and powershell(pwsh) for windows

Verf:
Windows system has two shell, Cmd and PowerShell (pwsh). 

Cmd can uses :: to comment, and  & to split. 
Powershell uses the # to comment, and ; to split, similar as bash. 

Toggleterm now distinguishes between linux and windows , but not deal with the cmd and powershell under the windows environment, resulting in the following error when using powershell:

This PR distinguishes the difference between cmd and powershell in windows environment to use the correct separator in powershell.

### DIFF
--- a/lua/toggleterm/terminal.lua
+++ b/lua/toggleterm/terminal.lua
@@ -10,8 +10,9 @@ local fmt = string.format
 local fn = vim.fn
 
 local is_windows = fn.has("win32") == 1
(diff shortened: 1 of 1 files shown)


akinsho:
Thanks for the contribution and the explanation @Verf I didn't know about this at all 👍🏿

Issue #48, opened 2021-06-04
Question: How to toggle view with all the terminals?


LoydAndrew:
Thanks for the great plugin. 
I have a question is there a way to have a command that toggles view with all the terminals that were created?
For example:
I created terminal with CTRl-\ than I created another terminal with a :ToggleTerm 2 
which leads to this view

![Screenshot_2021-06-04_12-44-39](https://user-images.githubusercontent.com/28489428/120782978-3bab0600-c533-11eb-9102-8a358e7dfa0f.png)

Than I want to hide all terminals (`:ToggleTerm all`  alike) and continue to work on editing files and after some time to toggle it again and have same view as before - 2 terminals bellow editing window (cause now CTRL -\ will only toggle one of the terminals, not all of them). 
This behaviour is similar to tmux prefix-z command which toggles maximize one pane. Or something like toggle terminal in heavy IDE-s like vscode, idea etc.

akinsho:
@LoydAndrew that's a cool idea tbh I hadn't really considered it ever. There isn't really a way to do that currently but I like the idea. I think a reverse command would be good as well i.e. toggle all with none open opens all the previously opened and then using that again toggles them all.

It wouldn't be too hard to implement basically could add a check for the `all` argument to the toggle term commannd and then loop through all terminals and call toggle. I guess the potentially weird edge case would be if one terminal was open and you hit `TermToggle all` should that open the rest or close the open one.

Maybe in that case it should work more explicitly `ToggleTermCloseAll` or `ToggleTermOpenAll` and then it can be explicitly mapped to opening any that aren't open or just closing them all.

Anyway I'd be happy to help direct a PR for this, most of the logic to do this is already there just needs commands defined and cobble thing together

```vim
" in toggleterm.vim
command! ToggleTermCloseAll lua require('toggleterm').toggle_all('close')
command! ToggleTermOpenAll lua require('toggleterm').toggle_all('open')
```
Then in `toggleterm.lua`

```lua
function M.toggle_all(command)
  for _, term in get_all() then
    if direction == "close" then
      term:close()
    else
      if not term:is_open() then
       term:toggle()
      end
    end
  end
end
```

akinsho:
@LoydAndrew it was pretty straightforward to do i.e. more or less exactly what I had above so I just did it in my lunch break. Can you test it out and see if it works

LoydAndrew:
@akinsho 
[...]

PR #107, opened 2021-10-30
feat: add single toggle all command

jorgerojas26:
This is useful i think for toggling terminals with just one mapping.

### DIFF
--- a/plugin/toggleterm.vim
+++ b/plugin/toggleterm.vim
@@ -11,5 +11,6 @@ endif
 " Count is 0 by default
 command! -count -complete=shellcmd -nargs=* TermExec lua require'toggleterm'.exec_command(<q-args>, <count>)
 command! -count -nargs=* ToggleTerm lua require'toggleterm'.toggle_command(<q-args>, <count>)
-command! ToggleTermCloseAll lua require'toggleterm'.toggle_all("close")
-command! ToggleTermOpenAll lua require'toggleterm'.toggle_all("open")
+" command! ToggleTermCloseAll lua require'toggleterm'.toggle_all("close")
+" command! ToggleTermOpenAll lua require'toggleterm'.toggle_all("open")
+command! ToggleTermToggleAll lua require'toggleterm'.toggle_all_terms()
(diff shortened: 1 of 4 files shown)


akinsho:
@jorgerojas26 thanks for the PR, this came up a while ago, and I added a `ToggleTermOpenAll` and a `ToggleTermCloseAll` command in #48 but the issue that came up there that lead to two commands was how to behave it there was one of several terminals opened. Should the toggle close the open one or open all the remaining hidden ones. This approach seems to choose to close the terminal if any is open and re-open all. 

Tbh I don't really mind but I don't want both implementations since it's unnecessary complexity so I'm in favour of choosing your approach, but it would be good to remove the `OpenAll` and `CloseAll` commands and just add a `ToggleTermToggleAll` command. Then add that to the documentation as well so users can find it.

jorgerojas26:
Sorry for late response, i have made the changes you requested, please check and tell me if they are ok for merging.

plugin/toggleterm.vim
```diff
@@ -11,5 +11,6 @@ endif
 " Count is 0 by default
 command! -count -complete=shellcmd -nargs=* TermExec lua require'toggleterm'.exec_command(<q-args>, <count>)
 command! -count -nargs=* ToggleTerm lua require'toggleterm'.toggle_command(<q-args>, <count>)
-command! ToggleTermCloseAll lua require'toggleterm'.toggle_all("close")
-command! ToggleTermOpenAll lua require'toggleterm'.toggle_all("open")
+" command! ToggleTermCloseAll lua require'toggleterm'.toggle_all("close")
```
akinsho, line 14: @jorgerojas26 I was thinking that this and all the other code related to the old functionality should be deleted not just commented out, can you remove the commented code here and also the function they reference e.g. `toggle_all` should be replaced with your implementation so there is only one function doing this and only one command

akinsho (changes requested):
@jorgerojas26 thanks for tweaking this, I think what would be best though would be to completely remove the old toggle_all code as mentioned in my comment and replace the implementation of `toggle_all` with `toggle_all_terms`

akinsho:
Just tested this out and it works nicely 👍🏿 . Thanks for making the changes and contributing 🚀

frt:
I think it would be nice if the open_mapping mapping did the ToggleTermToggleAll instead of the ToggleTerm. Or if configuration to do this was in the documentation.

akinsho:
@frt this issue is almost 2 years old so no longer relevant at all. FWIW the most recent commits of the plugin will now open and close the entire terminal view by default, so if you have 2 terminals it will close two and if you open it will open 2. So nothing in this issue relevant anymore.

frt:
Thank you @akinsho, I was using a tag (v2.2.1) without the most recent commits. After changing the Packer configuration to use one of the last commits of the main branch it is now working like I've expected in my first comment. Thank you.

PR #113, opened 2021-11-17
feat: Force close terminals via ToggleTermToggleAll

lcrockett:
See https://github.com/akinsho/toggleterm.nvim/commit/de9a2ef88e18e109744b94ea230bcc26c7232ffa

### DIFF
--- a/lua/toggleterm.lua
+++ b/lua/toggleterm.lua
@@ -222,16 +222,23 @@ end
 -- If any terminal is open it will be closed
 -- If no terminal exists it will do nothing
 -- If any terminal exists but is not open it will be open
-function M.toggle_all()
+function M.toggle_all(force)
   local ui = require("toggleterm.ui")
   local terminals = terms.get_all()
-  if not ui.find_open_windows() then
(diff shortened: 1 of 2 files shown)


lua/toggleterm.lua
```diff
@@ -222,16 +222,23 @@ end
 -- If any terminal is open it will be closed
 -- If no terminal exists it will do nothing
 -- If any terminal exists but is not open it will be open
-function M.toggle_all()
+function M.toggle_all(force)
   local ui = require("toggleterm.ui")
   local terminals = terms.get_all()
-  if not ui.find_open_windows() then
+
+  if force == true and ui.find_open_windows() then
```
akinsho, line 229:
```suggestion
  if force and ui.find_open_windows() then
```

akinsho (changes requested):
LGTM just made one minor style comment I'd rather see changed if possible.

lua/toggleterm.lua
lcrockett: You're right, it's redundant and gone. Thanks for the sharp look.

Issue #130, opened 2021-12-18
On_open for open_mapping


sarmong:
Can I set `on_open` option for the terminal that gets created with `open_mapping` ?

akinsho:
@sarmong this is already what the on open function does when you open *any* terminal it will use the default on open function

sarmong:
How do I set the default `on_open` function? I don't see such an option in `setup`

From the readme I see that `on_open` can only be passed to `Terminal:new` instance.

akinsho:
Ah apologies @sarmong I thought I also passed down an on open from the config. That can be added without too much work by adding an on_open key to `config.lua` which is initially set to null then where the terminal is created we can pass in the config's on open function or the one specific to that terminal if it was created ad hoc. Happy to help guide a PR if you're willing to raise one.

sarmong:
Yes, I'll open a pr. This option needs to be passed here?
https://github.com/sarmong/toggleterm.nvim/blob/265bbff68fbb8b2a5fb011272ec469850254ec9f/lua/toggleterm.lua#L116

akinsho:
@sarmong I wouldn't add it there. Instead you can add it when the terminal is created here
https://github.com/akinsho/toggleterm.nvim/blob/265bbff68fbb8b2a5fb011272ec469850254ec9f/lua/toggleterm/terminal.lua#L131-L153

If you add in a new line getting the value first from the terminal itself if not then from the config e.g.
```lua
term.on_open = term.on_open or conf.on_open
```

akinsho:
closed by #131

PR #131, opened 2021-12-20
feat: pass on_open option from setup

sarmong:
#130


akinsho:
Thanks for the PR @sarmong sorry for the delay, holidays etc.

PR #137, opened 2022-01-05
feat: pass on_close option from setup

tristone13th:
Similar to #131


akinsho (changes requested):
@tristone13th thanks for the PR. I prefer the way that the previous PR handles adding terminal config in the `Terminal:new` function rather than ad hoc inside other handlers like close. Can you move this code to where the other PR put the `on_open` function.

tristone13th:
> 

Updated, Thanks for your review.

akinsho:
Thanks for updating the PR @tristone13th, merged 👍🏿

PR #144, opened 2022-01-13
fix: some auto commands don't fire on Windows

tristone13th:


akinsho:
@tristone13th thanks for the PR can you verify whether this breaks things in linux and mac? If you have either system to hand. I'm a bit busy atm so can't test this out myself right now.

tristone13th:
> @tristone13th thanks for the PR can you verify whether this breaks things in linux and mac? If you have either system to hand. I'm a bit busy atm so can't test this out myself right now.

I tested it with a few commands under Ubuntu 20.04, it seems that everything is OK. I don"t have a mac, so sorry I cannot verify this.

akinsho:
Awesome thanks for checking 👍🏿

PR #149, opened 2022-01-24
feat: add terminal_mappings option

### LINKED ISSUE #122: [Feature request] add terminal_mappings
Means you can use the mapping to open and close a terminal when `insert_mappings` is set to false. My mapping uses my leader key (space) which was causing issues in insert mode. But if the mapping could work in terminal mode it would be great. It's a tiny change, happy to raise a PR if it makes sense?

MageJohn:
This adds a separate `terminal_mappings` option that when enabled (the default) maps keys to toggle the terminal in terminal mode. This is instead of applying those mappings when the `insert_mappings`  option is enabled.

Implements #122.


akinsho:
Thanks for the PR @MageJohn this looks great, will give it a go soon but looks good to merge on a brief skim

PR #157, opened 2022-02-16
fix: use proper newline characters on windows

### LINKED ISSUE #110: [Bug] TermExec passes null instead of string on Windows
When using `termexec cmd='anything'` it passes `null` to the terminal instead of `anything` and it doesn't even run it:

ezgif-5-78b0f6af5e25

I am on windows, and it the same thing if I am using `cmd` or `pwsh`.

celediel:
Should fix #110, by replacing \n with \r\n for cmd or \r for powershell. Make sure that `set shellslash` is executed on startup or `null` will still be sent to the terminal. Tested and working on Windows 11 with shell set as Windows' internal versions of cmd.exe and powershell.exe, and Powershell Core (pwsh.exe). I also verified in WSL that this doesn't break any Linux functionality.

### DIFF
--- a/lua/toggleterm/terminal.lua
+++ b/lua/toggleterm/terminal.lua
@@ -11,8 +11,10 @@ local fn = vim.fn
 
 local is_windows = fn.has("win32") == 1
 local is_cmd = string.find(vim.o.shell, "cmd")
+local is_pwsh = string.find(vim.o.shell, "pwsh") or string.find(vim.o.shell, "powershell")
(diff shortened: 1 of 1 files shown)


akinsho:
Awesome, thanks for raising this PR @celediel it's much appreciated

kraxli:
Unfortunately, it is not working for me on Windows 10 (neither with powershell nor with cmd) . So i still need to press `<enter>` after command is sent to toggleterm-terminal.

PR #159, opened 2022-02-19
fix: update error message to explicitly state compatible versions

superDross:
Explicitly state compatible versions of neovim in error message.


akinsho:
@superDross thanks for the PR 👍🏿

PR #163, opened 2022-02-26
fix: create terminal with called ID if not found

### LINKED ISSUE #146: <count>TermExec doesn't send command to the specified term, but opens a new term everytime
Hi, first thanks for the great plugin.

The issue I encountered are very simple. As I understand, if I specify a term count (say 5), then no matter I use `5TermExec cmd="ls" direction="horizontal"` or `lua require"toggleterm".exec("ls", 5, nil, nil, "horizontal")`, I should be sending the `ls` command to the same terminal with a count of 5. But each time I issue any of the commands above, it will just open a new term for me.

Again, thanks for the good work.

Daxtorim:
When using `5ToggleTerm`, but there is no terminal with id=5, a new terminal was created, *but not with id=5*. Instead a new terminal was created by calling `next_id()`, creating the weird behavior where `5ToggleTerm` would create new terminals until `next_id()` would finally return 5.

This PR fixes this by creating a new terminal with the specified ID when such a terminal does not already exist.

 fixes #146

### DIFF
--- a/lua/toggleterm/terminal.lua
+++ b/lua/toggleterm/terminal.lua
@@ -394,7 +394,7 @@ function M.get_or_create_term(num, dir, direction)
   if term then
     return term, false
   end
-  return Terminal:new({ dir = dir, direction = direction }), true
+  return Terminal:new({ id = num, dir = dir, direction = direction }), true
 end
 
 ---Get a single terminal by id, unless it is hidden


akinsho:
@Daxtorim I'd ignore the test error for now, If you check the output you'll see that ll the tests actually passed, but the process exits with the wrong exit code. Something in my pipeline seems to have changed, and I haven't figured out yet. Not sure if it's plenary or neovim or something I did yet.

akinsho (approved):
This also LGTM, one thing that I dimly recall looking at this was that initially there had been an issue of populating the terminal out of order 🤔 the current way worked around. I really can't remember what the issue was at the time. I'll try this out personally for the next couple of days just to see if anything strange occurs, otherwise this is good to merge.

PR #166, opened 2022-03-02
feat(highlights): add custom highlights

aspeddro:
This PR add support to highlights

```lua
local Term = require('toggleterm.terminal').Terminal:new {
  highlights = {
    Normal = 'Normal',
    EndOfBuffer = 'EndOfBuffer',
    VertSplit = 'VertSplit',
    StatusLine = 'StatusLine',
    StatusLineNC = 'StatusLineNC',
    SignColumn = 'SignColumn',
  },
}
```

### DIFF
--- a/lua/toggleterm/ui.lua
+++ b/lua/toggleterm/ui.lua
@@ -77,7 +77,7 @@ end
 ---apply highlights to a terminal
 ---@param term Terminal
 function M.hl_term(term)
-  local highlights
+  local highlights = {}
   local config = require("toggleterm.config")
   if term and term:is_float() or M.is_float() then
     local opts = term and term.float_opts or config.get("float_opts") or {}
@@ -85,6 +85,10 @@ function M.hl_term(term)
       fmt("NormalFloat:%s", opts.highlights.background),
       fmt("FloatBorder:%s", opts.highlights.border),
     }
+  elseif not vim.tbl_isempty(term.highlights) then
+    for key, value in pairs(term.highlights) do
+      highlights[#highlights + 1] = key .. ":" .. value
+    end
   elseif config.get("shade_terminals") then
     highlights = {
       "Normal:DarkenedPanel",
--- a/README.md
+++ b/README.md
@@ -302,6 +302,7 @@ Terminal:new {
   direction = string -- the layout for the terminal, same as the main config options
   dir = string -- the directory for the terminal
   close_on_exit = bool -- close the terminal window when the process exits
+  highlights = table -- a table with highlights (float terminal not supported)
   on_open = fun(t: Terminal) -- function to run when the terminal opens
   on_close = fun(t: Terminal) -- function to run when the terminal closes
   -- callbacks for processing the output
--- a/lua/toggleterm/terminal.lua
+++ b/lua/toggleterm/terminal.lua
@@ -146,6 +146,14 @@ function Terminal:new(term)
   term.direction = term.direction == "window" and "tab" or term.direction
   term.id = id or next_id()
   term.hidden = term.hidden or false
+  term.highlights = vim.tbl_deep_extend("keep", term.highlights or {}, {
+    Normal = "ToggleTermNormal",
+    EndOfBuffer = "ToggleTermEndOfBuffer",
+    VertSplit = "ToggleTermVertSplit",
+    StatusLine = "ToggleTermStatusLine",
+    StatusLineNC = "ToggleTermStatusLineNC",
+    SignColumn = "ToggleTermSignColumn",
+  })
   term.float_opts = vim.tbl_deep_extend("keep", term.float_opts or {}, conf.float_opts)
   term.on_open = term.on_open or conf.on_open
   term.on_close = term.on_close or conf.on_close
--- a/plugin/toggleterm.vim
+++ b/plugin/toggleterm.vim
@@ -5,6 +5,14 @@ if !has('nvim-0.5')
   finish
 endif
 
+" ToggleTerm Highlights
+highlight default link ToggleTermNormal Normal
+highlight default link ToggleTermEndOfBuffer EndOfBuffer
+highlight default link ToggleTermVertSplit VertSplit
+highlight default link ToggleTermStatusLine StatusLine
+highlight default link ToggleTermStatusLineNC StatusLineNC
+highlight default link ToggleTermSignColumn SignColumn
+
 "--------------------------------------------------------------------------------
 " Commands
 "--------------------------------------------------------------------------------


akinsho:
@aspeddro sorry for the radio silence on this PR just been busy with a bunch of things it's next on my list to review though

aspeddro:
@akinsho, alright. I'm making some adjustments on highlights.

lua/toggleterm/config.lua
```diff
@@ -17,10 +20,34 @@ local config = {
...
-      border = "Normal",
-    },
+  },
+}
+
+config.highlights = {
+  Normal = {
+    guibg = colors.shade_color(colors.get_hex("Normal", "bg"), constants.shading_amount),
+  },
+  NormalFloat = {
```
akinsho, line 30: @aspeddro can these be `snake_case` rather than PascalCase, not sure why they are if it's because of converting the names, so they can be used in a map then you can use
```lua
  -- convert 'highlight_value' to 'HighlightValue' -> snake to pascal
  local formatted = name:gsub("_(.)", name.upper):gsub("^%l", string.upper)
```
I think this would look much tidier if the casing was consistent

lua/toggleterm/ui.lua
```diff
@@ -77,31 +77,29 @@ end
...
   end
+
+  local highlights = vim.tbl_map(function(hl_group_name)
+    local hi_def = constants.highlight_group_name_prefix .. term.id .. hl_group_name
+    local hi_target = hl_group_name .. ":" .. hi_def
+
+    local hi_attrs = term.highlights[hl_group_name]
+
+    if hi_attrs.link then
+      vim.cmd(fmt("highlight default link %s %s", hi_def, hi_attrs.link))
```
akinsho, line 91: There is actually a private module you can currently use for highlighting in nvim called `vim.highlight` it has the methods `vim.highlight.create` and `vim.highlight.link` I think those have nicer semantics than  constructing the string hl commands yourself. It's a temporary solution anyway, since once 0.7 is out I'll migrate this to `nvim_set_hl`. But for now I think this will be better as
```lua
vim.highlight.create('Name', {guibg = 'Color', gui = 'GUI'})
```

lua/toggleterm/config.lua
```diff
@@ -17,10 +20,34 @@ local config = {
...
+    guibg = colors.get_hex("NormalFloat", "bg"),
+  },
+  FloatBorder = {
+    guifg = colors.get_hex("FloatBorder", "fg"),
+    guibg = colors.get_hex("FloatBorder", "bg"),
+  },
+  EndOfBuffer = {
+    guibg = colors.shade_color(colors.get_hex("Normal", "bg"), constants.shading_amount),
```
akinsho, line 38: It looks like a lot of these colour variables are repeated, so it would be more efficient to fetch them once at the top of the function and re-use them rather than calling `colors.get_hex` a bunch of times with the same arguments.
```lua
local normal_bg = colors.get_hex(BLAH)

-- ...
guibg = colors.shade_color(normal_bg)
```

lua/toggleterm/config.lua
```diff
@@ -17,10 +20,34 @@ local config = {
...
+  },
+  StatusLine = {
+    gui = "NONE",
+    guibg = colors.shade_color(colors.get_hex("Normal", "bg"), constants.shading_amount),
```
akinsho, line 42: The shade colour pattern is repeated a bunch, so you could create a small helper that does this automatically, e.g.
```lua
local function shade(color)
  return colors.shade(color, constants.shading_amount)
end
```

akinsho (changes requested):
@aspeddro thanks for the PR had a read through and have made some comments on tweaks that I think would be good

lua/toggleterm/config.lua
aspeddro: @akinsho keeping PascalCase makes mapping easier. One problem with the conversion is that it doesn't work for `StatusLineNC`, you would have to use `status_line_n_c`.

Another facility, passing the arguments to `vim.highlights.create` is made simpler.

lua/toggleterm/ui.lua
```diff
@@ -77,27 +77,26 @@ end
...
+    if hi_attrs.link then
+      vim.highlight.link(hi_def, hi_attrs.link)
+    else
+      vim.highlight.create(hi_def, hi_attrs)
+    end
+
+    return hi_target
+  end, vim.tbl_keys(term.highlights))
+
+  vim.api.nvim_win_set_option(term.window, "winhighlight", table.concat(highlights, ","))
```
aspeddro, line 99: moved to window api

lua/toggleterm/colors.lua
```diff
@@ -51,22 +58,39 @@ function M.color_is_bright(hex)
...
-  local bg_color = fn.synIDattr(fn.hlID("Normal"), "bg")
-  local darkened_bg = shade_color(bg_color, amount)
-  vim.cmd("highlight DarkenedPanel guibg=" .. darkened_bg)
-  vim.cmd("highlight DarkenedStatusline gui=NONE guibg=" .. darkened_bg)
+  local bg_color = M.get_hex("Normal", "bg")
+  local darkened_bg = M.shade_color(bg_color, amount)
+
+  local hl_group_name = constants.highlight_group_name_prefix
+
+  vim.cmd(fmt("highlight %s guibg=%s", hl_group_name .. "Normal", darkened_bg))
```
akinsho, line 87: This should also use highlight create

akinsho:
@aspeddro thanks for your work on this 👍🏿. I made a couple of tweaks to the documentation and some formatting but perhaps most significantly I made sure only `FloatBorder` and `NormalFloat` are set when highlighting a floating window and specifically set the value of those windows to `Normal` since this change would be disruptive to users otherwise as they might suddenly find their terminal looking different

akinsho (approved):
LGTM

will:
I switch between light mode and dark mode depending on the time of day (the rose-pine theme if it matters). Neovim handles this automatically at startup by doing some magic to detect the background of the shell and setting background to dark or light automatically. 

I did a manual bisect and before this patch, when I'd open a terminal, the background would always just match. However with this patch I get a dark background always.

Can you please point me in the right direction so I can always get a matching background?

akinsho:
@will the difference since this change is that rather than depending directly on X highlight, the plugin creates its own and derives them at startup from the theme. If you'd like to make a specific highlight follow another e.g. keep the background of the terminal matching the `Normal` background then you can use the solution for a similar issue I mentioned in https://github.com/akinsho/toggleterm.nvim/issues/196#issuecomment-1091666268

will:
Thanks @akinsho for the tip! 

It ended up being for me:

`highlights = { NormalFloat = {  link = "NormalFloat" }, FloatBorder = {link = "FloatBorder"}},
` 

to get it mostly working. Still need to do some tweaking I think, but now not super wrong.

Just for the future if anyone else comes across these comments: 
I was able to find those keys by looking at `lua print(vim.inspect(require("toggleterm.config").get("highlights")))` and seeing what was being used after all of the plugin setup/initialization

PR #186, opened 2022-03-22
feat: dynamic resize float terms

### LINKED ISSUE #168: [Feature] update term(float) size on VimResized autocmd
I have this config
```lua
require("toggleterm").setup({
    open_mapping = "<a-t>",
    direction = "float",
    float_opts = {
        border = config.border,
        winblend = 0,
        width = function()
            return math.ceil(vim.o.columns * 0.85)
        end,
        height = function()
            return math.ceil(vim.o.lines * 0.85)
        end,
        highlights = {
            border = "FloatBorder",
            background = "NormalFloat",
        },
    },
})
```
and when I open a term and resize neovim, tha terms keeps the same size

lopi-py:
closes #168

### DIFF
--- a/lua/toggleterm/terminal.lua
+++ b/lua/toggleterm/terminal.lua
@@ -99,6 +99,11 @@ local function setup_buffer_autocommands(term)
       fmt("<buffer=%d>", term.bufnr),
       fmt('lua require"toggleterm.terminal".delete(%d)', term.id),
     },
+    term:is_float() and {
+      "VimResized",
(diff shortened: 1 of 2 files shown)


akinsho (approved):
LGTM thanks @lopi-py

smangels:
How could I activate that feature. I'm using the latest version of toggleterm inside Astronvim and I had to define functions for width and height. Otherwise lazyvim inside toggleterm is not adjusted when going from half-size screen to fullscreen.

PR #182, opened 2022-03-21
fix: use specified shell on windows

rikuma-t:
Fix newline_chr,command_sep,comment_sep to be set correctly even if vim.o.shell and the specified shell are different on windows.

PR #191, opened 2022-04-01
Fix separator on pwsh

Shatur:
On Windows it should always be `&`.


Shatur:
Have issues as pointed in #182.

PR #192, opened 2022-04-02
fix: separator on windows with powershell

rikuma-t:
Solves the problem pointed in https://github.com/akinsho/toggleterm.nvim/pull/182#issuecomment-1086192923

Since command_sep and comment_sep are used to invoke the shell for toggleterm, it is correct to use `vim.o.shell`.
newline_chr is sent to the shell for toggleterm that has been started, so it is correct to use the specified shell.


akinsho:
@rikuma-t @Shatur thanks for debugging and looking into this, much appreciated. If you're both good with this solution will merge it shortly 👍🏿

PR #201, opened 2022-04-11
fix(send_lines): place cursor back in the original window

YuanYuYuan:
Thanks for this awesome plugin first! This PR tries to fix the bug of the cursor jumping back problem since the function `nvim_win_set_cursor` simply sets the cursor position on that window without jumping back to it.


akinsho (approved):
LGTM thanks @YuanYuYuan

PR #207, opened 2022-04-14
docs(README): rename 'window' direction option to 'tab'

GregoryBai:
The 'window' was deprecated in favor of 'tab'


akinsho (approved):
Thanks @GregoryBai much appreciated 👍🏿

PR #218, opened 2022-04-29
feat: add api to spawn tasks without opening a window

jedrzejboczar:
I was trying to spawn multiple terminal jobs in the background, but it seems that without `task:open()` the jobs won't be started. It is possible to do `task:open(); task:close()`, at least with `direction='float'`, but doing this for multiple terminals at once seems wrong.

This PR adds `task:spawn()` which only creates terminal buffer with the job, without opening a window yet. The window can be opened later and it doesn't seem to cause any problems. Using `nvim_buf_call` it was possible to make the `termopen` call actually use the background buffer instead of the current one.

### DIFF
--- a/lua/toggleterm/terminal.lua
+++ b/lua/toggleterm/terminal.lua
@@ -362,7 +362,7 @@ end
 
 ---Spawn terminal background job in a buffer without a window
 function Terminal:spawn()
-  if fn.bufexists(self.bufnr) == 0 then
+  if not (self.bufnr and api.nvim_buf_is_valid(self.bufnr)) then
     self.bufnr = api.nvim_create_buf(false, false)
     self:__add()
     api.nvim_buf_call(self.bufnr, function()
@@ -383,7 +383,7 @@ function Terminal:open(size, direction, is_new)
   if direction then
     self:change_direction(direction)
   end
-  if fn.bufexists(self.bufnr) == 0 then
+  if not (self.bufnr and api.nvim_buf_is_valid(self.bufnr)) then
     local ok, err = pcall(opener, size, self)
     if not ok then
       return utils.notify(err, "error")
--- a/lua/toggleterm/terminal.lua
+++ b/lua/toggleterm/terminal.lua
@@ -123,7 +123,10 @@ local function setup_buffer_autocommands(term)
   }
 
   if conf.start_in_insert then
-    vim.cmd("startinsert")
+    -- Avoid entering insert mode when spawning terminal in the background
+    if term.window == api.nvim_get_current_win() then
+      vim.cmd("startinsert")
+    end
     table.insert(commands, {
       "BufEnter",
       fmt("<buffer=%d>", term.bufnr),
--- a/lua/toggleterm/terminal.lua
+++ b/lua/toggleterm/terminal.lua
@@ -360,6 +360,19 @@ local function opener(size, term)
   end
 end
 
+---Spawn terminal background job in a buffer without a window
+function Terminal:spawn()
+  if fn.bufexists(self.bufnr) == 0 then
+    self.bufnr = vim.api.nvim_create_buf(false, false)
+    self:__add()
+    vim.api.nvim_buf_call(self.bufnr, function()
+      self:__spawn()
+    end)
+    setup_buffer_autocommands(self)
+    setup_buffer_mappings(self.bufnr)
+  end
+end
+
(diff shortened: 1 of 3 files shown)


akinsho:
@jedrzejboczar this largely looks alright to me, I just haven't had a chance to run it locally yet as I'm quite busy but will get to it soon.

jedrzejboczar:
Sure, no problem. Regarding CI, I'm not sure why it fails, running tests locally succeeds.

lua/toggleterm/terminal.lua
```diff
@@ -360,6 +360,19 @@ local function opener(size, term)
   end
 end
 
+---Spawn terminal background job in a buffer without a window
+function Terminal:spawn()
+  if fn.bufexists(self.bufnr) == 0 then
+    self.bufnr = vim.api.nvim_create_buf(false, false)
```
akinsho, line 366: So looking a bit at this now, UI code show generally be in the `ui` module, i.e. all window/buffer related stuff should live in there e.g. `ui.create_buf({ open = false})` or something.

lua/toggleterm/terminal.lua
```diff
@@ -360,6 +360,19 @@ local function opener(size, term)
   end
 end
 
+---Spawn terminal background job in a buffer without a window
+function Terminal:spawn()
+  if fn.bufexists(self.bufnr) == 0 then
```
akinsho, line 365: is this equivalent to `api.nvim_buf_is_valid(self.bufnr)` if so it would be good to avoid the `lua->viml` round trip via `vim.fn`

akinsho (commented):
Minor comments re. api usage and factoring some simple logic out of this file.

lua/toggleterm/terminal.lua
jedrzejboczar: I was using the same condition as in https://github.com/akinsho/toggleterm.nvim/blob/6c7f5dbdd69bc5611a85194ddca83ac2c8ee84d6/lua/toggleterm/terminal.lua#L373 but I can change both if you want.

lua/toggleterm/terminal.lua
jedrzejboczar: This is basically the same code as in [Terminal:open()](https://github.com/akinsho/toggleterm.nvim/blob/6c7f5dbdd69bc5611a85194ddca83ac2c8ee84d6/lua/toggleterm/terminal.lua#L373-L381), with the addition of `nvim_create_buf`, so shouldn't it rather stay nearby? Or do you mean refactoring both?

akinsho: Sorry I just meant the function call that creates a buffer should use the UI modules instead so there is a strict separation of concerns. The spawn function is fine where it is

jedrzejboczar: I moved buffer creation to ui module with: https://github.com/akinsho/toggleterm.nvim/pull/218/commits/dcc2660fcb48eb989066c91a33d365dead45911c

akinsho:
Thanks for making the tweaks @jedrzejboczar 👍🏾 one behaviour I'm noticing is that if a user has `start_in_insert` set to true then calling `term:spawn()` during startup (might be me misusing how this was intended) will cause nvim to open with your first buffer in insert mode. 

e.g. in `init.lua`

```lua
require('toggleterm').setup {
  start_in_insert = true
}
local t = Terminal:new({ cmd = 'bash' })
t:spawn()
```
Looking at the code seems to be because I call `startinsert` inside set buffer autocommands so it takes effect immediately and an autocommand so it happens on subsequent entries but even when moving this call to `nvim_buf_call` it still seems to affect the original non terminal buffer that was loaded.

jedrzejboczar:
Checking if the terminal has a window (and it is the current one) fixes the issue: https://github.com/akinsho/toggleterm.nvim/pull/218/commits/bba7f85f0a907297b328f67990f6fca19fb917dd

jedrzejboczar:
Done. I also rewritten the sentences:

> This will create a new terminal that runs the specified command once toggled if the `hidden` key is set to true. This terminal will not be toggled by normal toggleterm commands such as `:ToggleTerm` or the open mapping.

because they are wrong. The command won't run whatever the value of `hidden` is - looks like punctuation in the 2 sentences have been swapped.

akinsho:
Thanks for your work on this @jedrzejboczar just merged it 👍🏾

PR #221, opened 2022-05-05
Option to pass env/clear_env to jobstart

jedrzejboczar:
This allows to pass environmental variables to terminal jobs. 
While adding `env` test I also included an unrelated test for the new `term:spawn()` (can move it to separate PR if that's a problem).


akinsho:
Thanks for the PR LGTM 👍🏾

PR #224, opened 2022-05-11
fix: shade_color

dr666m1:
Thanks for this awesome plugin first!
Sometimes `shade_color()` returns invalid color and fails to darken window.

For example, my colorsheme is [terafox](https://github.com/EdenEast/nightfox.nvim) and its background color is `'#152528'`.
In this case `shade_color('#152528', -30)` returns `'#e191c'` (not `'#0e191c'`).

Zero padding fixes this problem.


akinsho (approved):
LGTM Thanks @dr666m1

PR #229, opened 2022-05-21
feat(toggleterm): add a simple winbar to show terminal data

akinsho:
This is currently blocked by a weird bug where the `winhighlight` for the `WinBar` does not get applied correctly. Possibly due to https://github.com/neovim/neovim/issues/18628

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -14,11 +14,11 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - name: Get neovim v0.7.0
+      - name: Get neovim nightly
         uses: actions/cache@v2
         with:
-          path: build/neovim/v0.7.0
-          key: ${{ runner.os }}-appimage-0.7.0
+          path: build/neovim/nightly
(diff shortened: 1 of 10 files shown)


famiu:
The bug is fixed now in the latest Neovim master

akinsho:
@famiu FYI I discovered that the issue I'm seeing is due to a different problem, I just realised that you can't set the `winhighlight` of the `Winbar` e.g. `setlocal winhighlight=Winbar:Number` won't work i.e. it will error saying it is invalid however `setlocal winhighlight=Normal:Number` does work.

famiu:
> @famiu FYI I discovered that the issue I'm seeing is due to a different problem, I just realised that you can't set the `winhighlight` of the `Winbar` e.g. `setlocal winhighlight=Winbar:Number` won't work i.e. it will error saying it is invalid however `setlocal winhighlight=Normal:Number` does work.

Ah, I don't think I can help you with that. `winhighlight` currently is problematic with a lot of things. Hopefully it'll be fixed soon.

akinsho:
image

Issue #223, opened 2022-05-09
Update colorsscheme and background of fly


Liverm0r:
When I set colorscheme with Telescope or `:colorscheme <name>`, theme is not applied to a hidden Toggle-term window. If toggle-window is opened, theme will be applied, but after toggling it the previous theme will reappear in the toggle-term window.

Currently I fixed it with applying `:on_open` function, where I check current theme and set it. Don't know if it's a bug. 

My config looks like (it's fennel, but it's readable I guess):

```fennel
(module plugin.toggleterm
  {autoload {nvim aniseed.nvim
             term toggleterm
             t    toggleterm.terminal}})

(defn update-bg []
  (set nvim.o.background nvim.o.background)
  (vim.api.nvim_command (.. "colorscheme " nvim.g.colors_name)))


(term.setup 
  {:size 25
   :open_mapping "•" ; alt + 8 (cmd + 8 from karabinner)
   :hide_numbers true
   :shade_terminals false
   :shading_factor 2
   :on_open update-bg
   ;:start_in_insert false
   :insert_mappings true
   :terminal_mappings true
   :direction :horizontal
   :persist_size true})

```

akinsho:
fixed by #231

PR #231, opened 2022-05-22
fix(colors): apply terminal highlights more consistently

### LINKED ISSUE #230: Plugin isn't setting or using NormalFloat highlight
I've been trying for some time to get the background color of the floating terminals to match the colors of the split ones, without success. I tried to investigate in the code and it seems to me that the plugin isn't setting the floating window to use the NormalFloat highlight, but I don't know Lua enough to be 100% sure of that.

I'd send some pics, but the images in the readme already illustrate the problem.

I tried manually setting the color in the setup() function, but it also doesn't work. These colors can't bee seen anywhere.

q_003

akinsho:
Also fix the issue of user highlights never being actually taken into account, and update the users highlights when the colorscheme changes.

fixes #230 #223 

@paulo-santana @Liverm0r can you both please test this PR

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,7 +158,7 @@ require("toggleterm").setup{
       guibg = <VALUE-HERE>,
     },
   },
-  shade_terminals = true,
+  shade_terminals = true, -- NOTE: this option takes priority over highlights specified so if you specify Normal highlights you should set this to false
   shading_factor = '<number>', -- the degree by which to darken to terminal colour, default: 1 for dark backgrounds, 3 for light
   start_in_insert = true,
   insert_mappings = true, -- whether or not the open mapping applies in insert mode
--- a/doc/toggleterm.txt
+++ b/doc/toggleterm.txt
@@ -147,7 +147,7 @@ show what options are available. It is not written be used as is.
           guibg = <VALUE-HERE>,
         },
       },
-      shade_terminals = true,
+      shade_terminals = true, -- NOTE: this option takes priority over highlights specified so if you specify Normal highlights you should set this to false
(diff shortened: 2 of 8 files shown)


paulo-santana:
Now setting the highlights in the setup works fine. The 'link' config still doesn't.
q_005

Liverm0r:
works, thanks!

akinsho:
@paulo-santana can you try the link key again, I wasn't passing this on before

paulo-santana:
@akinsho Now every ToggleTerm highlights have stopped working xD

q_007

akinsho:
@paulo-santana that's very strange and doesn't sound right, can you confirm how you are setting highlights since they shouldn't all break just because links were specified. Unless you are doing some linking of your own

paulo-santana:
This is how I'm setting them up
q_008

Does this happen only with me?

paulo-santana:
As a clarification, I tried disabling all other plugins that could mess with the highlights, the issue continues.

I discovered the plugin doesn't set the highlights until the first terminal is open. After that, the highlights appear on the listing, but the `ToggleTerm1NormalFloat` and `ToggleTerm1FloatBorder` are linking back to global Normal.
q_009


I tried setting the plugin's normal to red, and the split terminal does get the color. However, the floating ones don't.
q_010

akinsho:
@paulo-santana this should now be fixed this has all gotten quite complex, so I'll likely reject almost all future PRs related to highlighting. Trying to accommodate user feature requests has led to way too much complexity. Hopefully it resolves both your issues and I can call it a day.

paulo-santana:
@akinsho yup, it works fine now. Thank you for your work!

PR #234, opened 2022-06-02
feat(terms): delete internal term resources on shutdown.

asmodeus812:
Add delete call in the shutdown method for terminal, this will allow the terminal to properly be cleaned up when shutdown. Shutdown is used when
one desires to kill the terminal and not simply close it but rather
completely destroy it and free up a slot for another term instance.

### DIFF
--- a/lua/toggleterm/terminal.lua
+++ b/lua/toggleterm/terminal.lua
@@ -256,6 +256,7 @@ function Terminal:shutdown()
     self:close()
   end
   ui.delete_buf(self)
+  delete(self.id)
 end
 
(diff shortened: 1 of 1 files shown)


lua/toggleterm/terminal.lua
```diff
@@ -256,6 +256,7 @@ function Terminal:shutdown()
     self:close()
   end
   ui.delete_buf(self)
+  delete(self.id)
```
asmodeus812, line 259: This should not affect anything since shutdown is not called by anything but the test cases which should pass.

PR #239, opened 2022-06-05
feat: update send visual selection

etjupm:
Update the ToggleTermSendVisualSelection command by changing the "visual_selection" logic in M.send_lines_to_terminal in lua/toggleterm.lua.
For selection_type == "visual_selection", the lines now come from the new local function get_visual_selection which gets the lines based on the visual mode used to make the selection.


akinsho:
Seems good to me from what testing I did @etjupm 👍🏿 thanks for the PR

PR #240, opened 2022-06-10
fix(next_id): Terminal IDs should be unique.

nesqi:
Simplified next_id() by removing the 'ids' table and calling
get_all() instead. The previous implementation did not always
produce unique IDs.

Remodelled the next_id test case by creating actuall terminals.

Before this patch, if you created 3 terminals with ID 1, 2, 3
and then closed the first terminal with id 1 using "exit",
the next_id was calculated to 2, resulting in a duplicate ID.

### DIFF
--- a/lua/toggleterm/terminal.lua
+++ b/lua/toggleterm/terminal.lua
@@ -59,33 +59,19 @@ local terminals = {}
 --- @field on_close fun(term:Terminal)
 local Terminal = {}
 
----@type number[]
-local ids = {}
-
 ---@private
 --- Get the next available id based on the next number in the sequence that
 --- hasn't already been allocated e.g. in a list of {1,2,5,6} the next id should
 --- be 3 then 4 then 7
 ---@return integer
 local function next_id()
-  local next_to_use = #ids + 1
-  local next_index = #ids + 1
-  for index, id in ipairs(ids) do
-    if id ~= index then
-      next_to_use = index
-      next_index = index
+  local all = M.get_all(true)
+  for index, term in pairs(all) do
+    if index ~= term.id then
+      return index
     end
   end
-  table.insert(ids, next_index, next_to_use)
-  return next_to_use
-end
-
---- remove the passed id from the list of available ids
----@param num number
-local function decrement_id(num)
-  ids = vim.tbl_filter(function(id)
-    return id ~= num
-  end, ids)
+  return #all+1
 end
 
 ---Get an opened (valid) toggle terminal by id, defaults to the first opened
@@ -123,7 +109,6 @@ end
 --- @param num string
 local function delete(num)
   if terminals[num] then
-    decrement_id(num)
     terminals[num] = nil
   end
 end
@@ -511,13 +496,6 @@ if _G.IS_TEST then
     for _, term in pairs(terminals) do
       term:shutdown()
     end
-    ids = {}
-  end
-
-  ---@private
-  ---@param tbl number[]
-  function M.__set_ids(tbl)
-    ids = tbl
   end
 
   M.__next_id = next_id
--- a/tests/terminal_spec.lua
+++ b/tests/terminal_spec.lua
@@ -45,13 +45,22 @@ describe("ToggleTerm tests:", function()
     end)
 
     it("should assign the next id filling in any missing gaps", function()
-      t.__set_ids({ 1, 2, 5 })
-      local id = t.__next_id()
-      assert.equal(id, 3)
-      id = t.__next_id()
-      assert.equal(id, 4)
-      id = t.__next_id()
-      assert.equal(id, 6)
+      Terminal:new({ id = 2}):toggle() --2
+      Terminal:new():toggle() --1
+      Terminal:new():toggle() --3
+      Terminal:new():toggle() --4
+      Terminal:new({ id = 6 }):toggle() --6
+      local terms = get_all()
+      terms[3]:shutdown()
+      terms[1]:shutdown()
+      local new1 = Terminal:new():toggle()
+      assert.equal(1, new1.id)
+      local new3 = Terminal:new():toggle()
+      assert.equal(3, new3.id)
+      local new5 = Terminal:new():toggle()
+      assert.equal(5, new5.id)
+      local new7 = Terminal:new():toggle()
+      assert.equal(7, new7.id)
     end)
 
     it("should get terminals as a list", function()


akinsho:
Thanks for the PR @nesqi, I'll take a look shortly.

lua/toggleterm/terminal.lua
```diff
@@ -59,33 +59,19 @@ local terminals = {}
...
 local function next_id()
-  local next_to_use = #ids + 1
-  local next_index = #ids + 1
-  for index, id in ipairs(ids) do
-    if id ~= index then
-      next_to_use = index
-      next_index = index
+  local all = M.get_all(true)
+  for index, term in pairs(all) do
+    if index ~= term.id then
```
akinsho, line 70: I'm not sure what's happening here? What are these spots where there is a terminal, but the index is not the terminal's ID? Surely this means the terminal is in the wrong place somehow

nesqi: I'm not sure what the exact problem is. I had problems and solved them by simplifying the code.

The first problem I had was that the **ids** table wasn't in sync with the actual tabs. This was because decrement_id wasn't called when a tab was deleted. It turned out that this was fixed in a later version when I did a uplift of my fix but I still don't see the point of having double bookkeeping of the IDs, both in **ids** and the **terminals** table. There will always be a risk that it goes out of sync so, IF it's not needed, then maybe it's better to not have a separate table?

The second problem that IDs where reused might have to do with the ids table not being sorted? M.get_all() sorts all the IDs, so you know that if, index ~= id, then there must be a gap. In an unsorted table this is not true, it could just be out of order.

As for the updated tests it now captures the use-case I had problems with. I'm pretty sure the test will not pass without the updates to next_id.

nesqi: Or maybe your question was why `index ~= term.id`, sometimes is true?
This happens when you terminate a terminal. So if you have terminals with IDs 1, 2, 3 and terminate terminal 1 you end up with terminals with IDs 2, 3. It's also possible to create terminals with specific IDs, that is another way to create a gap.

akinsho: @nesqi thanks for explaining the question was definitely more around the specific comparison and how it works. I guessed that `index ~= term.id` was intended to identify gaps but what is confusing is that if there is a gap I'd expect the table to look like
```lua
{
  [1] = Term{...},
  [2] = <empty>,
  [3] = Term{...},
}
```
In this case 2 is a gap from a deleted terminal, so I would expect the check to see if that id is available to look like
```lua
for index, term in pairs(terms) do
  if index and not term then -- I expect term to be nil not term.id
    return index
  end
end
```

Basically my question is why is `term.id` possible, there should be no terminal there if there is a gap? Is there some lingering terminal object still there? This would indicate that the deletion of terminals does not remove them from the list of all terminals correctly

nesqi: That is how **terms** is constructed in M.get_all(). There will never be any index-gaps.

```
function M.get_all(include_hidden)                                                                                                                             
  local result = {}                                                                                                                                            
  for _, v in pairs(terminals) do                                                                                                                              
    if include_hidden or (not include_hidden and not v.hidden) then                                                                                            
      table.insert(result, v)                                                                                                                                  
    end                                                                                                                                                        
  end                                                                                                                                                          
  table.sort(result, function(a, b)                                                                                                                            
    return a.id < b.id                                                                                                                                         
  end)                                                                                                                                                         
  return result                                                                                                                                                
end    
```

akinsho: @nesqi sorry I think I'm not being clear thanks for pointing that out, I remember painfully writing that function to avoid gaps, I guess what I'm trying to ask is, since as you mention `get_all` doesn't have gaps so the list is always full then why when searching through it does it check the id and if it it doesn't match returns it's id.

I'm asking that if the list is full and there are no gaps *what does your logic do?*. Essentially can you explain how this works, I know I wrote the logic in this file, but it's been a while and I can't understand how this new change works, if the list has no gaps pairs will iterate through all keys and return the value, we've just confirmed that there will always be a value for each key, so there is always a terminal at a spot as we iterate so if you just check ID to see if it matches and re-assign it, what happens to the terminal that was in that spot.

nesqi: Ok.

So, if there are NO gaps (among the IDs, the indexes never have any gaps as discussed) the table will look like `{1=1, 2=2, 3=3, ...}`. In that case the 'if' will never trigger and the second 'return' will run, 'return #all+1'.

If the IDs in use aren't in sequence the table might look like '{1=1, 2=3, 3=4}', (if you have created 4 terminals and shut down the second one) in that case, the 'if' will trigger in the second go-around and return 2 as the next_id to be used.

Since M.get_all() returns a table with no gaps in its indexes, it is equivalent to do 'for i, t in pairs(all)' and 'for i, t in ipairs(all)'. Maybe ipairs would have been more readable.

So, the for loop holds the count from 1 up the the number of terminals in 'index', it then compares 'index' with 'term.id' to see if 'all' is missing a term with ID of 'index'. Since 'all' is sorted, we know that if there is a miss we have found the lowest available free id, and we can return that.

akinsho: @nesqi thanks for going through this 👍🏿 that all makes sense 💯 .
So what I was vaguely concerned about but couldn't piece together was how this works around terminals getting pushed out of the table. In the example you show with `2=3` if 2 is then returned as the next available ID, note this function is used in `Terminal:new` which eventually adds a terminal in via The `__add` function. 
```lua
---Add a terminal to the list of terminals
function Terminal:__add()
  if not terminals[self.id] then
    terminals[self.id] = self
  end
  return self
end
```
The problem that this will introduce is that it will check to see if `terminals` has that ID, in this case it will be the `2` that was returned from `next_id` which means the `if not terminals[self.id] then` will be false and so this terminal will not be added to the list of terminals.

I think to fix this and stop terminals getting orphaned, that block should instead use `table.insert(terminals, self.id, self)` since insert allows adding an element to a list and adjusting the positions

nesqi: That shouldn't be a problem.

`terminals` and the table returned from `M.get_all()` are different tables. `terminals` has gaps while the other does not.

If `terminals` holds 3 terminals with IDs 1, 3 and 4 then it looks like,
```
terminals[1] = Term{id=1,...}
terminals[2] = nil
terminals[3] = Term{id=3,...}
terminals[4] = Term{id=4,...}
```
Then if `all = M.get_all()` is called it will return a table that looks like,
```
all[1] = Term{id=1,...}
all[2] = Term{id=3,...}
all[3] = Term{id=4,...}
```
next_id will return 2, and the new term will be placed in `terminals[2]`.

akinsho:
@nesqi thanks for working on this, going through it and updating the tests 👍🏿

nesqi:
No problem! @akinsho thank you for writing the plugin!

PR #245, opened 2022-06-16
feat: persist terminal mode

### LINKED ISSUE #76: Track whether in insert mode or not during toggle
Hi, thanks for making this plugin! It is awesome!

I like the `start_in_insert = true` feature a lot. However, sometimes (for example, when I am looking at output of a lot of tests), I use `C-\ C-n` to exit insert mode and scroll up, and toggle to get to the test file in question. However, if I want to refer to the test output again, now when I toggle back to the terminal it enters insert mode and goes back to the terminal input and I have to scroll back again. 

It would be nice if the plugin can track whether the user is in insert mode or not and choose _not_ to enter insert mode if a user has `C-\ C-n` in the terminal. 

One way I think this could be achieved is by using autocommands that work only in the terminal. Thoughts?

akinsho:
This PR allows each terminal to keep track of the mode the user was last in and return to in when they re-open the terminal. Fixes #76

@kdheepak can you give this a try? it should keep track of the terminal state and re-use whatever it last was

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,14 +148,14 @@ require("toggleterm").setup{
     -- highlights which map to a highlight group name and a table of it's values
     -- NOTE: this is only a subset of values, any group placed here will be set for the terminal window split
     Normal = {
-      guibg = <VALUE-HERE>,
+      guibg = "<VALUE-HERE>",
     },
     NormalFloat = {
       link = 'Normal'
     },
     FloatBorder = {
-      guifg = <VALUE-HERE>,
-      guibg = <VALUE-HERE>,
+      guifg = "<VALUE-HERE>",
+      guibg = "<VALUE-HERE>",
     },
   },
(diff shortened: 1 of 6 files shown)


Shatur:
@akinsho I had the same problem as the author of #76, so the feature sounds interesting. But is there a way to switch from Toggleterm to another buffer without leaving insert mode?
Currently I use mapping like `<C-\\><C-N><C-w>h` which changes my mode to `normal`.

akinsho:
@Shatur I'm not sure, there used to be a feature like this ages ago that a user contributed, but there weren't any tests for it and I think maybe it regressed at some point without anyone commenting

Shatur:
Got it. I probably could "hack it" by settings the mode to insert manually in my mapping to be able to switch between windows without messing the mode.

PR #254, opened 2022-06-29
docs(README): update `vim-plug` installation instructions

gh4n:
The current installation instructions yield a syntax error. Fixed `vim-plug` installation line to match the examples in the [vim-plug docs](https://github.com/junegunn/vim-plug).


akinsho (approved):
Thanks for the PR @gh4n LGTM

PR #264, opened 2022-07-12
fix: Ensure terminals are deleted correctly when multiple open

morgsmccauley:
When a terminal is opened an `autocmd` is attached to the given terminal buffer to delete the terminal from the in-memory register. This becomes a problem when multiple terminals are opened as previously created auto commands are replaced by subsequent creations, meaning only the last terminals reference will be removed when it is closed, all others will remain.

Rather than relying on multiple `autocmd`s which have closure over the terminals they need to close, this PR uses a single command which uses the buffer number from the callback to locate the terminal which needs to be deleted. My neovim/lua knowledge isn't the best so feel free to edit if you have a better solution.

My motivation for wanting this behaviour is that I use sessions to switch between projects. If I have multiple terminals open within one project/session, when switching sessions, all terminal buffers will be closed but not all of them will be removed from memory. Then, when I open a new terminal in the new session, the previous terminal references are used and therefore retain the previous directory.

### DIFF
--- a/lua/toggleterm/terminal.lua
+++ b/lua/toggleterm/terminal.lua
@@ -9,7 +9,6 @@ local config = lazy.require("toggleterm.config")
 local utils = lazy.require("toggleterm.utils")
 ---@module "toggleterm.constants"
 local constants = lazy.require("toggleterm.constants")
-local AUGROUP = "ToggleTermBuffer"
 
 local api = vim.api
 local fmt = string.format
@@ -20,6 +19,8 @@ local mode = {
   NORMAL = "n",
   UNSUPPORTED = "?",
 }
+
+local AUGROUP = api.nvim_create_augroup("ToggleTermBuffer", { clear = true })
 
 local is_windows = fn.has("win32") == 1
 local function is_cmd(shell) return string.find(shell, "cmd") end
@@ -119,7 +120,6 @@ end
 ---@param term Terminal
 local function setup_buffer_autocommands(term)
   local conf = config.get()
-  api.nvim_create_augroup(AUGROUP, { clear = true })
   api.nvim_create_autocmd("TermClose", {
     buffer = term.bufnr,
     group = AUGROUP,


lua/toggleterm/terminal.lua
```diff
@@ -121,9 +132,11 @@ local function setup_buffer_autocommands(term)
   local conf = config.get()
   api.nvim_create_augroup(AUGROUP, { clear = true })
   api.nvim_create_autocmd("TermClose", {
-    buffer = term.bufnr,
```
akinsho, line 124: If this isn't local to a buffer then it should use the pattern for toggleterm buffers, you can find this where the other terminal autocommands are. It's something like `pattern = 'term://*toggleterm*'`

akinsho: It probably should also not be set in a `setup_buffer_autocommands` because this will recreate this autocommand every time this function is called, which is every time a toggleterm is created.

lua/toggleterm/terminal.lua
```diff
@@ -121,9 +132,11 @@ local function setup_buffer_autocommands(term)
...
     group = AUGROUP,
-    callback = function() delete(term.id) end,
+    callback = function(args)
+      local closedTerm = find_terminal_by_bufnr(args.buf)
```
akinsho, line 137: Snake case is the convention in this project and most lua projects (that I'm aware of) i.e. should be `closed_term`

akinsho (changes requested):
@morgsmccauley thanks for the PR I haven't really investigated if/why this is now necessary since previously before migrating to the new autocommands this I believed worked without issue.

I've requested a few changes nonetheless, and it looks generally ok, after those but will first want to look more closely at why a pretty old mechanism needs a new solution

akinsho:
> previously created auto commands are replaced by subsequent creations

This sounds like an/the underlying bug. This behaviour is not right. Buffer numbers are unique in a vim session and so autocommands for a buffer should not be being replaced by subsequent buffers

morgsmccauley:
@akinsho Dug a bit deeper on this and think I now see the problem. The `ToggleTermBuffer` augroup is being recreated on every terminal open, and since `clear = true` it wipes any previous autocommand that had been set up. I think there are two options:
1. Set `augroup` to `clear = false`
2. Create the `augroup` once, outside of `setup_buffer_autocommands`

Not sure of the side effects of either so will wait for your feedback before updating.

akinsho:
@morgsmccauley option two sounds like the right decision and an error on my part the augroup is essentially constant and shouldn't be in the function

morgsmccauley:
@akinsho sounds good - updated the PR 👍🏽

akinsho (approved):
LGTM 🚀

Issue #175, opened 2022-03-10
[Question] Term name


tinnedkarma:
Right now, the only way to tell terminals apart is using `count`.
Is this an explicit feature you wanted to have? Why not assign a name for each terminal instance?
I feel like I'm missing information when I'm switching between terminal instances (Yes, I can configure `statusline` to add info about my current terminal, but I usually focus my eyes on terminal and not the `statusline`). More so, I like the approach  of telescope.nvim of adding a title at top of each section (probably not called sections).
Would you consider this feature for implementation, or an pull request?

akinsho:
@LuchianMihai so I think this is doable but using the `b:term_title` variable. You can collect the user input using `vim.ui.input` then setting this variable for the terminal. The difficulty will be how to show this name to the user. I'd rather avoid creating complex UI just for this. Either way would be happy to review a PR for this feature if you're willing to work on it

PR #272, opened 2022-07-17
feat: add ToggleTermSetName command

akinsho:
This will allow a terminal to have its name set by the user, which then means it will use this name inside the winbar for that terminal


akinsho:
fixes #175

Issue #278, opened 2022-07-23
[Feature] dynamic width and height for float terminal


xiyaowong:
like `size` , we can pass function to `float_opts.width` `float_opts.height` :)

akinsho:
@xiyaowong please feel free to make a PR for this 🙏🏾 it's not high on my list as I have other things in currently working on

PR #279, opened 2022-07-24
doc: float terminal accepts a function for size

xiyaowong:
close: #278


xiyaowong:
After looking at the source code, because they all use the same function to handle the size, so the feature I needed was originally supported : )

akinsho:
I thought that was likely the case but didn't have time to check, thanks for the docs update

Issue #289, opened 2022-08-08
[Feature idea]: Alternate smart-toggle strategies


rish987:
Hi there! I've been using this plugin for quite a while (it finally got me off of tmux!), and wanted to ask for your opinion on this idea for a new configuration option + set of external utility functions. Currently, when using `:ToggleTerm` without a count, it defaults to "smart-toggling", which has the effect of opening the first unhidden terminal if none are open, and closing open terminals one by one otherwise. However, being a frequent user of `:tcd [directory]`, I would find it more useful for terminals to be opened on a per-tab basis, rather than per-session, and there are probably `:lcd [directory]` users out there who would like to have this effect per-window.

So, I'm proposing that we add a new config option `toggle_strategy` that can be set to `"first"`, `"by_tab"`, or `"by_window"` that lets the user select their preference. And we could expose the utility functions that work under the hood to find the correct terminal to toggle (based on context) so it is easy for the user to define additional commands if they want to toggle sometimes by tab, sometimes by window, sometimes globally, etc.

Of course, this is possible already with some amount of configuration work by the user, but I think explicitly supporting this could make things much easier on their end. And we can take the opportunity to generalize/unify some of the logic in the current implementation.

rish987:
Another smart-toggle helper we could add is one that opens the last-used terminal, which I think could be very useful to avoid having to provide a count before `<C-\>` when possible, though I think we might have to add some state to make this work. Anyways, I wanted to gauge your opinion on #290 before working on this (in a separate PR).

akinsho:
Hi @rish987 

Sorry I haven't spent much time to properly weigh in, I'm generally intrigued by the idea but also concerned about introducing complexity that increases the maintenance burden by either adding complexity to the plugins usage of its code. Ultimately whatever change is added would have to be as minimal as possible in the first instance and more importantly easy to maintain. Ultimately the change sounds good but if it means every future change now has the overhead of considering this I'd be less inclined.

rish987:
Hi, thanks for the input! This should be fairly self-contained I believe, I tried to keep it minimal in #290. Really all that changes is that instead of hard-coding the smart-toggle strategy, we've abstracted this to allow the user to pick one from a set of given strategies (and which defaults to the original "first non-hidden terminal" strategy). It doesn't touch any core aspects of the code, so any PRs not relating to smart-toggling specifically shouldn't have to even consider it.

[...]

PR #290, opened 2022-08-08
feat: context-based toggling

rish987:
Proof-of-concept for #289.


lua/toggleterm.lua
```diff
@@ -321,9 +293,56 @@ function _G.___toggleterm_winbar_click(id)
...
+      -- Re-open the first terminal toggled
+      return terms.get_toggled_id()
+    end
+    return M.get_last_open_term()
+  end,
+
+  -- use a terminal corresponding to the current tabpage handle
+  by_tabpage = function()
+    if not ui.find_open_windows() then
+      return vim.api.nvim_get_current_tabpage()
```
akinsho, line 329: I'm not sure I see how this works? Are you envisioning only one terminal per tab? This won't align with future plans to allow creating multiple terminals per tab

lua/toggleterm.lua
```diff
@@ -338,7 +357,17 @@ function M.toggle(count, size, dir, direction)
   if count >= 1 then
     toggle_nth_term(count, size, dir, direction)
   else
-    smart_toggle(count, size, dir, direction)
+    local toggle_strategy
+    if type(config.toggle_strategy) == "function" then
+      toggle_strategy = config.toggle_strategy
```
akinsho, line 362: I think before anything is exposed this way, i.e. we take a function it needs to be ironed out how this function will work so the api can remain consistent over time. Are there existing ideas of how a user will add this or is the general idea someone will think of something in which case I'd rather wait till there is a specific use case and cater to those rather than taking a function without a sense of if it's desirable by users and what such a function would take as arguments if any.

doc/toggleterm.txt
```diff
@@ -466,10 +475,20 @@ You can create your on commands by using the lua functions this plugin provides
...
+    local toggleterm = require'toggleterm'
+    local strategies = toggleterm.toggle_strategies
+    local cmd = vim.api.nvim_create_user_command
+
+    cmd("TermGitPush", function(opts) toggleterm.exec("git push", opts.count, 12) end, {count = 1})
+    cmd("TermGitPushF", function(opts) toggleterm.exec("git push -f", opts.count, 12) end, {count = 1})
+    cmd("ToggleTermTab", function(opts) toggleterm.toggle_command(opts.args, strategies["by_tabpage"]()) end, {})
 <
 
+The table `require'toggleterm'.toggle_strategies` provides a set of utility
```
akinsho, line 487: I'm very, very keen to keep the API of this function very lean and quiet straightforward, I really don't want to expose a bunch of foot guns or internal APIs that then lock me into maintaining functions in a certain way. I'd much rather we focus on clear use cases rather than maybe people will find them useful apis

akinsho (commented):
Added a couple of comments, primarily to streamline things and minimise where possible, as well as for clarifying some use cases

lua/toggleterm.lua
rish987: This essentially has exactly the same effect as `smart_toggle` had before, it's just that we're using the current tab's handle as the terminal ID rather than the ID of the first terminal found in the list of unhidden terminals. It doesn't restrict you to one terminal per tab or anything, it just defines the default behavior of `:ToggleTerm` without a count.

lua/toggleterm.lua
```diff
@@ -321,9 +293,56 @@ function _G.___toggleterm_winbar_click(id)
...
+  -- use a terminal corresponding to the current tabpage handle
+  by_tabpage = function()
+    if not ui.find_open_windows() then
+      return vim.api.nvim_get_current_tabpage()
+    end
+    return M.get_last_open_term()
+  end,
+
+  -- use a terminal corresponding to the current window handle
+  by_window = function()
```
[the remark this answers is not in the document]
rish987: Good point, removed.

lua/toggleterm.lua
rish987: Sounds good, I agree let's implement strategies on our own as requests come in. Removed.

doc/toggleterm.txt
rish987: The main thing here is that people (myself certainly included) may want to have the option of either smart-toggling by session or by tab, so I exposed this to allow for that. If we want to hide the strategies table altogether an alternative would be to define commands for each variant (e.g. :ToggleTermTab), I'm just not sure about the extent to which you're keen on potentially proliferating such commands variants as we get request for new smart-toggle strategies (at first glance, I personally don't have much against doing so).

lua/toggleterm.lua
```diff
@@ -321,9 +293,56 @@ function _G.___toggleterm_winbar_click(id)
...
+    end
+  end
+  if not target then
+    utils.notify("Couldn't find a terminal to close", "error")
+    return
+  end
+  return target.id
+end
+
+M.toggle_strategies = {
```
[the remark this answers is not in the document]
rish987: Yes that's cleaner, thanks. I went with your second approach.

rish987:
Hi @akinsho, thanks for the review! I'm marking this as a draft because I wanted to get your opinion on my recent idea in #289. Essentially, yes I agree that we should hide the strategies API, but I think that we should have the command `:ToggleTermLast` for opening the last opened terminal and `:ToggleTermTab` for opening a terminal specific to this tab. Whenever you open a terminal in a tab, it is pinned to that tab (via an internal map from tabpages to terminal numbers). And the default strategy (`:ToggleTerm`) is to open the pinned terminal corresponding to that tabpage if there is one, or the last opened terminal if there isn't. This should preserve existing behavior pretty well while allowing for very convenient fine-grained control over how terminals are toggled by default.

Again, this will all be very well-contained to defining the default behavior of `:ToggleTerm`. Let me know if this sounds like a reasonable approach and I will be happy to implement it!

doc/toggleterm.txt
akinsho: @rish987 so by default a user can add terminal to their session options which should persist all terminals in a neovim session and then toggleterm will resurrect them (think this might have regressed recently for some reason). I'm trying to ensure that the motivations for all these changes are super clear and that the code only add/fixes those specific use cases.

I'm fine with having a few strategy commands, I personally don't imagine these will grow a lot since tbh after several years of maintaining this plugin this is the first request for any alternative strategy so I don't imagine that will suddenly grow to dozens

rish987:
@akinsho I've overhauled this PR, removing strategies in the process. I added the "Smart-Toggling" section to the README explaining the new design, let me know what you think!

README.md
```diff
@@ -260,6 +260,22 @@ You can send commands to a terminal without opening its window by using the `ope
...
+
+When `:ToggleTerm` is used without a prefixed count, it defaults to "smart-toggling".
+This feature is designed to use your current neovim context to determine which terminal
+to open, thus helping to minimize the number of times you have to explicitly provide a count.
+
+Currently, terminal contexts exist on a per-tab basis. Each tab is "pinned" to
+the first terminal that is opened in it. If you call `:ToggleTerm` from a new tab,
+it defaults to opening, (and thus pinning) the last-opened terminal
+(which can be done explicitly with `:ToggleTermSmartLast`).
+Alternatively, you can call `:ToggleTermSmartNew` within a new tab to associate
```
akinsho, line 273: I think rather than a bunch of new commands which tbh I'm having a little trouble following the workflow for these it would be better if a user set a strategy for tabs or something, and then it just behaved like you intend without needing these extra commands

akinsho (commented):
@rish987 I've had another look over and made some comments, in general my biggest hurdle with this PR is I find it hard to envision the workflow this unlocks, it seems to implicitly work for a certain usage of toggleterm which is a bit alien to me

It seems quite focused on how toggleterm behaves in tabs specifically and trying to localise terminals to tabs? Is that right?

It would be good to get a clear description of the exact problem this solves (I don't object to adding in a strategy like this I just currently don't follow what it's for)

I think one reason for this is expanding the use of the term "smart" which I think ends up being quite subjective. I think maybe this change should be in terms of tabs like `ToggleTabTerm` or something

README.md
```diff
@@ -260,6 +260,22 @@ You can send commands to a terminal without opening its window by using the `ope
 
 see `:h expand()` for more details
 
+### Smart-Toggling
```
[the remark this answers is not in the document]
rish987: Thanks for bringing that up. I've done away with "smart toggle" in favor of "context-based toggle", that should hopefully be more clear.

README.md
```diff
@@ -260,6 +260,22 @@ You can send commands to a terminal without opening its window by using the `ope
...
+
+When `:ToggleTerm` is used without a prefixed count, it defaults to "smart-toggling".
+This feature is designed to use your current neovim context to determine which terminal
+to open, thus helping to minimize the number of times you have to explicitly provide a count.
+
+Currently, terminal contexts exist on a per-tab basis. Each tab is "pinned" to
```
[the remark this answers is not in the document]
rish987: At this point what it does is open the last-used terminal if there isn't anything pinned yet (see the updated README). I think this is a lot more intuitive that what we did before (open the first-opened terminal, whatever that was), do you agree?

README.md
rish987: The workflow I was envisioning was: I'm in a tab, I have a terminal associated with it. Now, when I open a new tab, I have a choice: either associate it with the same terminal (`:ToggleTerm`, which opens the last-used terminal if one isn't already associated) or a brand new one (`:ToggleTermNew`). Now, if my work in this new tab diverges to the point that I want to associate a different terminal with it, I can do `:ToggleTermContextClear` followed by `:ToggleTermNew` (or I can open some other existing terminal).

rish987:
@akinsho Sorry for the long wait, but I'm back now. I've taken your comments into consideration and done a refactor. The main difference is that we're now toggling w.r.t. a generic "context" rather than tying things to tabs. I've added the config option `get_context` where you can specify a function to return a context, which can be _anything_ -- tabpage, window, git root directory, etc. I've also added some tests. Let me know if you can relate at all to the workflow I've described above.

rish987:
Thanks for the feedback @akinsho. I've just added logic to default to the original behavior of `smart_toggle` when `get_ctx == nil` (which was to open the first-opened terminal in the list of currently active terminals). So it's no longer a requisite additional layer of complexity on top of `:ToggleTerm`, it's something you have to explicitly opt-into by providing `get_ctx`.

I'm not entirely sure what you had in mind w.r.t. API calls, did you want to leave `:ToggleTerm` alone and have a separate `:ToggleCtxTerm` command (or low-level API call) for this? I thought it best to just override the default "open first-opened terminal" behavior because it wasn't particularly useful for me at least -- this could also be a good opportunity for us to define what exactly we expect the default to be.

rish987:
Hey! I think I'lll close this one off, finally -- I haven't looked at this in a while, but I'm starting to think that it's best to keep this plugin minimal and keep any notion of context-based toggling limited to a gist that perhaps makes use of a general plugin for context detection (and plugs into the API that Toggleterm provides). I'll update you on any progress I make there.

PR #299, opened 2022-08-27
fix(terminal): keep focus in current on TermExec in 'float' orientation

bobokrut:
The value of go_back has been inverted on check. This caused the terminal to close when set to **false**

### DIFF
--- a/lua/toggleterm/terminal.lua
+++ b/lua/toggleterm/terminal.lua
@@ -290,7 +290,7 @@ function Terminal:send(cmd, go_back)
   cmd = type(cmd) == "table" and with_cr(unpack(cmd)) or with_cr(cmd)
   fn.chansend(self.job_id, cmd)
   self:scroll_bottom()
-  if not go_back then
+  if go_back then
     ui.goto_previous()
     ui.stopinsert()
   end
--- a/lua/toggleterm/terminal.lua
+++ b/lua/toggleterm/terminal.lua
@@ -283,6 +283,12 @@ function Terminal:scroll_bottom()
   if ui.term_has_open_win(self) then api.nvim_buf_call(self.bufnr, ui.scroll_to_bottom) end
 end
 
+function Terminal:is_focused() return self.window == api.nvim_get_current_win() end
+
+function Terminal:focus()
+  if ui.term_has_open_win then api.nvim_set_current_win(self.window) end
+end
+
 ---Send a command to a running terminal
 ---@param cmd string|string[]
 ---@param go_back boolean? whether or not to return to original window
@@ -290,9 +296,11 @@ function Terminal:send(cmd, go_back)
   cmd = type(cmd) == "table" and with_cr(unpack(cmd)) or with_cr(cmd)
   fn.chansend(self.job_id, cmd)
   self:scroll_bottom()
-  if go_back then
+  if go_back and self:is_focused() then
     ui.goto_previous()
     ui.stopinsert()
+  elseif not go_back and not self:is_focused() then
+    self:focus()
   end
 end
 


akinsho:
@bobokrut so I made this change on purpose because the call to scroll no longer actually causes toggleterm to change window, so it scrolls from the original window a user was in so if `go_back` is true there is no need to do anything else since you never left the window. You mention that something is closing instead, but none of this functionality should have anything to do with closing anything.

Can you describe the exact situation you are trying to fix

bobokrut:
@akinsho [Since you always set go_back to `false` for the "float" terminal](https://github.com/akinsho/toggleterm.nvim/blob/main/lua/toggleterm.lua#L175), now commands similar to `:TermExec cmd="echo hi" direction="float" size=40` are executed, the focus is returned immediately to the previous window, and this leads to closing of the terminal. Additionally, now go_back does the quite opposite of that is supposed to according to the docs. So, I thought you made a typo:)
> By default focus is returned to the original window after executing the command (except for floating terminals). Use argument go_back=0 to disable this behaviour.

akinsho:
@bobokrut I completely forgot that people would be using this in the context of floats, since I don't really use this plugin like that. Can you confirm thought that when used in the context of a split that go back works as intended because if so then the check should maybe be `not go_back or term:is_float()` or something? Regarding it doing the opposite, I think you mean in the case of floats it does the opposite maybe

bobokrut:
@akinsho If I understood the purpose of go_back option, it should return the cursor to the previous buffer when set to `1` and stay in the terminal when set to `0`. Correct me if I'm wrong. What is happening now is:
`:TermExec cmd="echo hi" direction="vertical" go_back=1 size=40`
https://user-images.githubusercontent.com/45918782/187070780-29ce8220-f70a-40f0-98e2-5540a8980835.mp4
`:TermExec cmd="echo hi" direction="vertical" go_back=0 size=40`
https://user-images.githubusercontent.com/45918782/187070894-7d644079-3cc5-4a7c-9c41-6e7b8bca55a4.mp4

So, with `not go_back`, when `go_back=1` the cursor stays in the terminal, while `go_back=0` changes focus to the previous buffer. This I meant when I was saying that it is doing the opposite thing

akinsho:
@bobokrut this is strange because I get the opposite behaviour to you. Can you share your toggleterm config? For me `go_back=1` keeps the cursor in the current buffer and 0 puts it in the terminal.

bobokrut:
@akinsho Sure! Here you go:) I don't remember I set something special and before this change everything was fine for me

```lua
local toggleterm_status_ok, toggleterm = pcall(require, "toggleterm")

if not toggleterm_status_ok then
    vim.notify("Toggleterm is not installed\nSee https://github.com/akinsho/toggleterm.nvim", vim.log.levels.WARN,
        { title = "Plugins" })
    return
end

toggleterm.setup {
    direction = "tab",
    shade_terminals = false,
    float_opts = {
        border = "curved"
    },
    on_open = function(term)
        vim.api.nvim_buf_set_keymap(term.bufnr, "t", "<Esc>", "<C-\\><C-n>", { silent = true })
    end
}
```

bobokrut:
@akinsho To me the logic of [this block](https://github.com/akinsho/toggleterm.nvim/blob/main/lua/toggleterm/terminal.lua#L293) of code is so. If `go_back=0`, which means that `go_back` in the function will be false, what is inside the `if` statement will be executed (the focus will be returned, and the mode will be back to normal) because there is `not` before the `go_back` variable. And this explains to me the behaviour that I'm getting. Sorry, if I'm miss something

akinsho:
@bobokrut I think the confusion is because how each of the functions in that block runs is not immediately clear, i.e. the scroll to bottom function use `nvim_win_call` which means that you can do something like the following
```lua
api.nvim_win_call(term.window, function()
	scroll_logic()
end)
```
The function above will not change the current window, focus will temporarily be moved to `term.window` but then neovim will automatically move it back because that's how `win_call` works. I refactored the `scroll_to_bottom` logic to use this. So actually by the time `scroll_to_bottom` is called. Focus will *still be in the original window,* so the reason I inverted the logic is that go back now happens implicitly. So the logic actually doesn't make sense to `go_back` any more it should be inverted since we are already in the starting point so really it should check if go back is false and focus the terminal or not e.g.

```lua
  if not go_back and not self:is_float() then api.nvim_set_current_win(self.window) end
```

akinsho:
Ok so I can reproduce the issue you are seeing, but it only happens *the first time the terminal is opened*

akinsho:
The following patch seems to fix the issue
```patch
diff --git a/lua/toggleterm/terminal.lua b/lua/toggleterm/terminal.lua
index 290031c..d78e426 100644
--- a/lua/toggleterm/terminal.lua
+++ b/lua/toggleterm/terminal.lua
@@ -283,6 +283,12 @@ function Terminal:scroll_bottom()
   if ui.term_has_open_win(self) then api.nvim_buf_call(self.bufnr, ui.scroll_to_bottom) end
 end
 
+function Terminal:is_focused() return self.window == api.nvim_get_current_win() end
+
+function Terminal:focus()
+  if ui.term_has_open_win then api.nvim_set_current_win(self.window) end
+end
+
 ---Send a command to a running terminal
 ---@param cmd string|string[]
 ---@param go_back boolean? whether or not to return to original window
@@ -290,9 +296,11 @@ function Terminal:send(cmd, go_back)
   cmd = type(cmd) == "table" and with_cr(unpack(cmd)) or with_cr(cmd)
   fn.chansend(self.job_id, cmd)
   self:scroll_bottom()
-  if not go_back then
+  if go_back and self:is_focused() then
     ui.goto_previous()
     ui.stopinsert()
+  elseif not go_back and not self:is_focused() then
+    self:focus()
   end
 end
```

Although this looks similar to your changes, one *key difference* is that in case the window never changed e.g. the terminal was already opened, and you run term exec with the terminal open go to previous would put you in the wrong place if you never moved.

bobokrut:
@akinsho Thanks! Now everything seems working properly

akinsho:
Merged 👍🏿 thanks for raising the PR @bobokrut

PR #301, opened 2022-09-01
style: add colon to terminal selection prompt

ObserverOfTime:
This is so [fzf-lua](https://github.com/ibhagwan/fzf-lua)'s selection menu can properly override the prompt suffix.

https://github.com/ibhagwan/fzf-lua/blob/0944e1e/lua/fzf-lua/providers/ui_select.lua#L90
```lua
    ['--prompt']          = prompt:gsub(":%s?$", "> "),
```


akinsho:
Not sure that I entirely understand, but don't mind having a colon there 🤷🏿‍♂️

Issue #141, opened 2022-01-11
"couldn't find a terminal to close" when toggling a split when a tab is open


ioreshnikov:
Hey!

I've noticed a mildly annoying bug: if you open a terminal with `:ToggleTerm tab`, switch back to the original tab and then call `:ToggleTerm split`, then you get the error message from the subject. It works perfectly fine in the other order -- a split terminal is being converted into a tab one, but not vice versa.

Best,
Ivan

akinsho:
Thanks for raising this, sorry lost sight of this issue. Just reproduced this locally and is definitely a bug but tbh it's not immediately clear what the right behaviour should be, but I guess in keeping with the general logic, if you haven't specified a terminal, and you decide to run `Toggleterm direction=tab` then move to another tab and convert it to a `horizontal` split then only one terminal should be manipulated i.e. the last one opened.

#309 attempts to fix this, but working with some many layouts is fiddly, so please test it out and let me know how it works for you. If you notice the winhighlight bleeding through feel free to mention, but frankly that's an upstream bug with a fix in nightly, but a hack in stable which I don't want to add here. Will patch it for nightly, but leave it as is for stable till 0.8 is released.

ioreshnikov:
Oh man, I completely forgot about this! :)

> it's not immediately clear what the right behaviour should be

I agree. I remember at the time skimming the code and thinking that this behaviour might be intentional. The use-case I have might be a bit different from what the extension initially intended – instead of multiple terminals I just want to have one. 

> keeping with the general logic, if you haven't specified a terminal, and you decide to run Toggleterm direction=tab then move to another tab and convert it to a horizontal split then only one terminal should be manipulated i.e. the last one opened.

This sounds good. I will give it a try today.

ioreshnikov:
@akinsho I have just tested it and it seems to be working as intended. Thanks!

akinsho:
Cool, I've just merged this in 👍🏿

ioreshnikov:
@akinsho I've just discovered a bug:

* open nvim
* call `ToggleTerm tab`
* switch back to the original empty tab
* call `ToggleTerm tab` one more time
* the terminal tab stays there, but you cannot switch to it

akinsho:
Will have a look 👍🏿

akinsho:
[...]

PR #309, opened 2022-09-09
fix(ui): check all tabs for current terminal's window

akinsho:
When seeing if the terminal is toggled open, but only check current win for the current terminals windows when finding windows to see if any terminals are open.


akinsho:
fixes #141

Issue #307, opened 2022-09-08
[Question/Bug] auto scroll not triggering on term open


asteroidalaz:
When I pass `auto_scroll = true` to the terminal, it is not autoscrolling in normal mode.

akinsho:
@asteroidalaz please provide specific examples of what exactly doesn't work what exact config you are using and what the expected outcome is, otherwise I will close this issue out for lack of information

asteroidalaz:
I am trying to open toggle term with this command
```lua
    local run_term = Terminal:new({
        cmd = my_command,
        direction = "horizontal",
        close_on_exit = false,
        auto_scroll = true
    })

    run_term:toggle()
```
but the output is not automatically scrolling, (I am in the normal mode in the terminal window).
I am always at the top of the line

akinsho:
@asteroidalaz can you give an example of the type of command you are running that isn't working, that is kind of important. The "scroll" just jumps to the end of the buffer, so the behaviour will kind of depend on what you are doing. It would be useful to have the entire steps that you are following so I can just do exactly what you are doing. If the command is sensitive then something similar will be good enough

akinsho:
This should now be fixed by #319 so going to close this out. Please report back if it's not

PR #319, opened 2022-09-18
fix(terminal): fix scroll_bottom (#307)

daschw:
Hi, first thanks for your awesome plugin!
I came across #307 today and I think this change fixes the issue (at least for my use case it does).


akinsho (approved):
LGTM 👍🏿 thanks for catching this bug and raising this issue

PR #326, opened 2022-10-01
Term splits respect splitbelow, splitright

bnwa:
Perhaps a further enhancement would be to define the `direction` enum as `top | left | right | botton | float` and update logic accordingly, but this seemed to be the smallest change

### DIFF
--- a/lua/toggleterm/ui.lua
+++ b/lua/toggleterm/ui.lua
@@ -202,17 +202,29 @@ function M.switch_buf(buf)
   if cur_buf ~= buf then vim.cmd(fmt("keepalt buffer %d", buf)) end
 end
 
-local split_commands = {
-  horizontal = {
-    existing = "rightbelow vsplit",
-    new = "botright split",
-    resize = "resize",
-  },
-  vertical = {
-    existing = "rightbelow split",
-    new = "botright vsplit",
-    resize = "vertical resize",
-  },
+local function split_command(direction, existing)
+  if 'horizontal' == direction then
+    if existing == true then
+      return"rightbelow vsplit"
+    elseif vim.o.splitbelow then
+      return "botright split"
+    else
+      return"topleft split"
+    end
+  elseif 'vertical' == direction then
+    if existing == true then
+      return "rightbelow split"
+    elseif vim.o.splitright == true then
+      return "botright vsplit"
+    else
+      return "topleft vsplit"
+    end
+  end
+end
+
+local resize_command = {
+  horizontal = "resize",
+  vertical = "vertical resize"
 }
 
 ---Guess whether or not the window is a horizontal or vertical split
@@ -273,17 +285,18 @@ end
 --- @param term Terminal
 function M.open_split(size, term)
   local has_open, win_ids = M.find_open_windows()
-  local commands = split_commands[term.direction]
 
   if has_open then
     -- we need to be in the terminal window most recently opened
     -- in order to split it
     local split_win = win_ids[#win_ids]
+    local command = split_command(term.direction, true)
     if config.persist_size then M.save_window_size(term.direction, split_win) end
     api.nvim_set_current_win(split_win)
-    vim.cmd(commands.existing)
+    vim.cmd(command)
   else
-    vim.cmd(commands.new)
+    local command = split_command(term.direction, false)
+    vim.cmd(command)
   end
 
   M.resize_split(term, size)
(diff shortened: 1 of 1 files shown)


bnwa:
The expanded enum would permit the scenario wherein a user wants her buffer splits along one direction of an axis but the term to appear in the opposite direction but that seems niche

lua/toggleterm/ui.lua
```diff
@@ -202,17 +202,29 @@ function M.switch_buf(buf)
...
-    new = "botright split",
-    resize = "resize",
-  },
-  vertical = {
-    existing = "rightbelow split",
-    new = "botright vsplit",
-    resize = "vertical resize",
-  },
+local function split_command(direction, existing)
+  if 'horizontal' == direction then
```
akinsho, line 206: It's clear to me if this condition is reversed, i.e.
```suggestion
  if direction == 'horizontal' then
```

lua/toggleterm/ui.lua
```diff
@@ -202,17 +202,29 @@ function M.switch_buf(buf)
...
-    existing = "rightbelow vsplit",
-    new = "botright split",
-    resize = "resize",
-  },
-  vertical = {
-    existing = "rightbelow split",
-    new = "botright vsplit",
-    resize = "vertical resize",
-  },
+local function split_command(direction, existing)
```
akinsho, line 205: `existing` is a confusing name for this variable, I think `is_open` or `win_open` is a clearer name

lua/toggleterm/ui.lua
```diff
@@ -202,17 +202,29 @@ function M.switch_buf(buf)
...
+  if 'horizontal' == direction then
+    if existing == true then
+      return"rightbelow vsplit"
+    elseif vim.o.splitbelow then
+      return "botright split"
+    else
+      return"topleft split"
```
akinsho, line 212: there should be a space between the return and the string here

lua/toggleterm/ui.lua
```diff
@@ -202,17 +202,29 @@ function M.switch_buf(buf)
...
-    resize = "resize",
-  },
-  vertical = {
-    existing = "rightbelow split",
-    new = "botright vsplit",
-    resize = "vertical resize",
-  },
+local function split_command(direction, existing)
+  if 'horizontal' == direction then
+    if existing == true then
```
akinsho, line 207: There is no need to directly compare against `true` this should just be
```suggestion
    if existing then
```
The same goes for all the other `variable == true` unless there is a specific need to worry about comparing with some other truthy value vs true, then just use `if variable then`

lua/toggleterm/ui.lua
```diff
@@ -273,17 +285,18 @@ end
...
     local split_win = win_ids[#win_ids]
+    local command = split_command(term.direction, true)
     if config.persist_size then M.save_window_size(term.direction, split_win) end
     api.nvim_set_current_win(split_win)
-    vim.cmd(commands.existing)
+    vim.cmd(command)
   else
-    vim.cmd(commands.new)
+    local command = split_command(term.direction, false)
+    vim.cmd(command)
```
akinsho, line 299: There is no need for an interim variable here
```suggestion
    vim.cmd(split_command(term.direction, false))
```

akinsho (changes requested):
@bnwa it would be good to have a clear description of what problem *exactly* you were facing that this PR solves. This plugin was initially designed to work exactly as it does to achieve the desired effect. I don't know how these changes will behave and whether they will be breaking for users who expect a specific kind of layout.
Would be good to get some clarification on exactly how these changes will impact how this plugin works

I've also made a few stylistic comments

bnwa:
> @bnwa it would be good to have a clear description of what problem _exactly_ you were facing that this PR solves. 

As a user, I'd like to be able to control the direction of term splits. The code in your repo hardcoded the direction of splits in the `split_commands` table. For myself, it isn't ergonomic in my vim workflow to shift whatever I'm editing to the left for the sake of the terminal buffer, particularly as I leave them open during a session.

As I saw it, there were two choices then:
1. A more fine-grained revision to the `direction` enum, perhaps `top | left | bottom | right | float`, but this would force all users to update their config  and vim package managers have no especial means of notifying users of breaking changes
2. Update the handling of `direction` in the UI to respect `splitbelow` and `splitright` options; these options are already the users way of communicating their preference of which direction to open a new split along one of the two axes of `direction`.

However, I see now that I'd wrongly recalled that the neovim defaults for these options run contrary to your chosen split directions, which would be a major disruption for extant users.

So perhaps then a third option presents itself: a new config field indicating which (actual) direction along one of the two available axes represented in the current `direction` enum, defaulting to the extant behavior on `main`, would be better and less disruptive

PR #337, opened 2022-11-03
feat: add desc property to key mappings

GnikDroy:
This property can be used in Neovim 0.7 and above. Useful for plugins like which-key.nvim


akinsho (approved):
LGTM 👍🏿

PR #185, opened 2022-03-22
feat: send line(s) to the terminal

gaborvecsei:
#172 

Purpose of the PR is to make "sending lines" from a file to the toggled terminal available for the user (as simple as possible). Can come handy in multiple situations (e.g. executing code with `ipython`)

The PR contains:
- Sending a single line where our cursor is `:ToggleTermSendCurrentLine <T_ID>`
- Sending the lines from a visual selection `:ToggleTermSendVisualLines <T_ID>`
- Sending the visual selection (e.g. block of code, single word) `:ToggleTermSendVisualSelection <T_ID>`

(`<T_ID>` is an optional argument. It's the ID of the terminal where we want to send the line(s) )

Example

toggleterm_test

PR #339, opened 2022-11-04
Support sending contents of current buffer to terminal

chancez:
Enhancement of #185 to support sending the entire buffer to the terminal.

### DIFF
--- a/lua/toggleterm.lua
+++ b/lua/toggleterm.lua
@@ -215,6 +215,9 @@ function M.send_lines_to_terminal(selection_type, trim_spaces, cmd_data)
     local res = utils.get_line_selection("visual")
     start_line, start_col = unpack(res.start_pos)
     lines = utils.get_visual_selection(res)
+  elseif selection_type == "current_buffer" then
+    start_line, start_col = unpack(api.nvim_win_get_cursor(0))
+    lines = api.nvim_buf_get_lines(0, 0, -1, false)
   end
 
   if not lines or not next(lines) then return end


akinsho:
I commented in #185, but I'd rather see a note with the workaround added to the README as @gaborvecsei recommended rather than add yet another command. I was on the fence about the original addition, one of the reasons was that I feared that exactly this sort of scenario would come up where even more commands were added. I'd much rather keep this plugin simple rather than add commands for every scenario.

chancez:
> I commented in #185, but I'd rather see a note with the workaround added to the README as @gaborvecsei recommended rather than add yet another command. I was on the fence about the original addition, one of the reasons was that I feared that exactly this sort of scenario would come up where even more commands were added. I'd much rather keep this plugin simple rather than add commands for every scenario.

Can we keep the current_buffer functionality at least? I’m fine with defining my own command but I’d rather not have to deal with handling logic around making it work based on the existing visual selection logic.

gaborvecsei:
@chancez Something like this you could put in your `nvim` config?
(I could not try it out, I'm not at my laptop)

```
vim.api.nvim_create_user_command("MyFn",                                                                                
    function(_)                                                                                                         
        vim.api.nvim_feedkeys("ggVG", "n", false)                                                                       
        vim.cmd("ToggleTermSendVisualLines")                                                                        
    end,                                                                                                                
    {}                                                                                                                                                                                                       
)
```

chancez:
> @chancez Something like this you could put in your `nvim` config? (I could not try it out, I'm not at my laptop)
> 
> ```lua
> vim.api.nvim_create_user_command("MyFn",                                                                                
>     function(_)                                                                                                         
>         vim.api.nvim_feedkeys("ggVG", "n", false)                                                                       
>         vim.cmd("ToggleTermSendVisualLines")                                                                        
>     end,                                                                                                                
>     {}                                                                                                                                                                                                       
> )
> ```

Having to use feed keys is a big hack IMO. Exposing the right knobs in the Lua functions is the "right" approach. Using feedkeys has some drawbacks like lag, it's not silent (you can see the movements), etc. 

I'd rather just re-implement `M.send_lines_to_terminal` instead, which is pretty annoying since I'd be just copying/pasting it with my modifications to support it.

gaborvecsei:
@chancez Not my call if this gets implemented or not, I am just the kind stranger who was trying to help in the quickest way possible

chancez:
Well, I've updated the PR to just update the function, without adding a new command, and defined my own command in my `init.lua` instead which uses the updated function. @akinsho hopefully this is acceptable?

akinsho:
Sorry, I'm on holiday atm and haven't/don't have the time to look at this in great detail. If I'm understanding correctly, the entire purpose here is to avoid having to select the buffer then sending it to terminal? This solution still seems to depend on yet another conditional, which I'm certain will just invite more down the line. I don't know why (without taking the time) someone can't select the whole buffer and then use send visual selection command, is it that doing this is too slow?

This seems even more obscure now, since the only value would be for someone who knows that this conditional is in place.

Please try to understand that in general what I want for this plugin is to provide generalized APIs/commands for people to do what they want rather than implementing every specific use case directly in this plugin. What argument would I have, having accepted this, against the next user who would like the visible content of a buffer sent to the terminal or the contents of window x sent.

chancez:
The purpose is to expose APIs that use native Lua APIs that don't rely on `ex` commands to manipulate the cursor, or enter commands, which will be faster, and provide better UX>

Based on your argument, it almost sounds like the original PR #185 shouldn't have ever been merged, since you could have implemented all that using Ex commands too. 

> I want for this plugin is to provide generalized APIs/commands 

Is that not what this PR is doing? Right not the `send_lines_to_terminal` API is very oriented around visual selection, yet vim supports many ways of targeting text. Requiring everything to be done in terms of a visual selection makes this API _less_ generalized. I don't see how extending it to support more selection methods is going against your goals.

akinsho:
@chancez definitely regret merging #185, although I very much appreciate the work of everyone who contributed to this. As I tried to make clear in the README this plugin is primarily about opening terminals it seemed like such a small innocuous snippet that adding it wouldn't/shouldn't cause too much harm, but unfortunately it has led to a great deal of conversation/issues on a feature I do not use and have very little specific interest in.

Send lines to terminal is a fine solution as is because this was never intended to start trying to support all the ways that neovim can handle text. Sending x number of lines or just the visual selection seems reasonable as a solution as it covers a lot of use cases and has done for a lot of people. It seems that this is largely around avoiding pressing ggVG? This is the part I'm struggling to understand. The specific need for this functionality, i.e. what problem are we trying to solve?

In either case, I will leave this till I'm back from holiday.

chancez:
>  is largely around avoiding pressing ggVG

Yes. My point is: we're in Lua land now, there's no reason to use ex commands when we have Lua APIs to directly express what we want. It's like the difference in using `vim.api.nvim_create_autocmd()` and `vim.cmd [[autocmd! ..... ]]`. Once you have proper Lua APIs, why would you use ex?

If you feel that the functionality of sending text to terminals doesn't belong in ToggleTerm, I think that's entirely valid, but given it's presence in the project, I assumed that it would make sense to extend it for more use-cases. 

Given that `M.exec` is exposed by ToggleTerm, I can simply copy/paste `M.send_lines_to_terminal` into my configuration with my changes and just not use the implementation in ToggleTerm, and then ToggleTerm doesn't need to handle it at all. Maybe this functionality could be moved into a different nvim plugin if it's not wanted in core ToggleTerm.

gacallea:
hi :)

FWIW I also would like to send a buffer/file content commad. I'm currently setting up Neovim for live coding and this feature would be very much beeficial.

Please and thank you

gacallea:
post scriptum: I found this plugin that is designed to send lines or files to a REPL. Haven't figured out how to customize yet. Not sure it can interact with toggleterm yet. given @akinsho regrets, it could be an opportunity to reverse teh merge and suggest a configuration or something in the docs. my 2c.

gacallea:
@chancez @akinsho I managed to get my head around IRON plugin. It works great. One can also customize mappings and functions as IRON exposes its guts to do just that. I hope it can help you both.

Issue #363, opened 2022-12-12
[Suggestion] Status for the float windows


Blackmorse:
Since winbar is disabled for float terminal windows, I'm really missing information about the terminal.

Will it make sense to have this status at the top of the float windows? Like it is at [floatterm](https://github.com/akinsho/toggleterm.nvim):

![image](https://user-images.githubusercontent.com/13163993/207051341-9f3293eb-c8df-4be4-9451-4dd60474794a.png)

akinsho:
@Blackmorse currently neovim 0.8.1 does not support floating window titles, but it will do in neovim 0.9.0. So currently this is not possible without a bunch of hacky workarounds which I definitely will not introduce but could be possible to add by anyone fairly easily once neovim 0.9 is released

PR #368, opened 2022-12-14
feat: enable switching to the next and previous float terminal

Blackmorse:
### TL;DR
Introducing of **ToggleTermFloatNext/ToggleTermFloatPrev** commands for the `float` terminals. Just switching to the next/previous terminal id from the list in a cycled manner.

### Motivation
Since I usually use `float` windows for the terminals, in my flow I'm missing the feature of "switching to the previous/next terminal".

E.g. It's common for me to have 3 terminals at the same time, and switching between them with the `count` feature is not very convenient, as I have to keep in mind some kind of mapping of the terminal number to its actual state. I'd like to switch between them just by going forward/backward

After Neovim 0.9 release and together with this feature (https://github.com/akinsho/toggleterm.nvim/issues/363) process of switching will become even more transparent for the user.

It's my first experience in Lua/Neovim development, so don't judge me too much, I will be happy to fix any notes and issues :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -294,6 +294,10 @@ You can map this to a key and call it with a count, which will then prompt you a
 Alternatively you can call it with just the name e.g. `:ToggleTermSetName work<CR>` this will the prompt you for which terminal it should apply to.
 Lastly you can call it without any arguments, and it will prompt you for which terminal it should apply to then prompt you for the name to use.
 
+### ToggleTermFloatNext/ToggleTermFloatPrev
+
+This functions allow you to switch the terminals by cycling through them when you are in a float one.
+
 ### Set terminal shading
 
 This plugin automatically shades terminal filetypes to be darker than other window
--- a/lua/toggleterm.lua
+++ b/lua/toggleterm.lua
@@ -70,6 +70,37 @@ local function smart_toggle(_, size, dir, direction)
   end
 end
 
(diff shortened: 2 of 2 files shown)


beem812:
I'd love to have this functionality available

akinsho (changes requested):
@Blackmorse thought I like this idea, I'm not sure if it should be specialized to floats alone. I don't want to have custom commands depending on the UI of a terminal, ideally. I think this should be a command that cycles between all terminals that have a matching layout e.g. if it's a split then this command should cycle through all terminals which are splits, if it is float then cycle through all floats

Bonus points would be a command like
`ToggleTermCycleNext direction=float`
and the default could be
`ToggleTermCycleNext direction=match`, this would cycle through all terminal with a matching direction
and without the direction argument, then
`ToggleTermCycleNext` would just cycle any non-hidden terminal regardless of direction

Blackmorse:
@akinsho , that for the feedback a lot

I like the way you want to unify this plugin, and it is the first thing I've tried to implement. However, non-floating terminals are quite different and I stuck with some ambiguous behaviour, and therefore implemented it only for floating windows which looks natural here.

It would be nice If you can give me an advice how it should work.

First question that I remember: how cycling should work for the hidden splits? In case split terminal is hidden, should it appear when cycling through? (For floating windows it's exactly the point)

PR #378, opened 2023-01-03
fix: add missing self:on_create() to Terminal:spawn()

jedrzejboczar:
Currently the `on_create` callback is not being called when creating a task using `Terminal:spawn()` instead of `Terminal:open()`.


akinsho (approved):
LGTM 👍🏿

PR #387, opened 2023-02-02
doc(README): fix vim plug installation instructions

TornaxO7:
I think that this PR would avoid problems like those:
image

[Here](https://vi.stackexchange.com/questions/39967/cannot-open-terminal-ini-nvim/39968?noredirect=1#comment73213_39968)'s the vim question.


akinsho (approved):
LGTM 👍🏿 thanks

Issue #359, opened 2022-12-06
Can ToggleTerm Respect Nvim Tree Window? 


DennisTheMenace780:
I have Toggle Term and Nvim Tree set up and what I find happening is that the window for NvimTree can be _above_ Toggle Term, or it can be to the left of Toggle Term if I do a toggle on NvimTree. I have included two image links below to demonstrate that.

I need to figure out a way to get Toggle Term to respect the boundary of the Nvimtree window when I open Toggle Term. Otherwise what happens in the case that Nvimtree is above Toggle Term, if I do a window motion from my main buffer into Toggle Term, and then from Toggle Term to the main buffer, I actually put my cursor on NvimTree because it is directly above Toggle Term.

Edit: I believe that I can figure this out, however I need to be able to specify the dimensions of the terminal window in horizontal mode. When I look for how to do this, it seems that there is only a `size` option, and then from there it's not clear to me how the horizontal width of the terminal is being set. If I can figure out how to set the horizontal width of the terminal, then I think I got this sorted out!

[What I don't want](https://freeimage.host/i/HCPWNqB) arises when I first activate the NvimTree window, and then bring up Toggle Term. 

[What I do want](https://freeimage.host/i/HCPWjsV) arises when I first activate the ToggleTerm window, and then bring up NvimTree.

akinsho:
@DennisTheMenace780 I get this issue and variations of it *a lot* I've also explained in some detail the reasoning for this #97 is one example. I will leave it to you to have read through historical issues.

Apologies in advance because this won't be of much help, but I'm really trying to manage the amount of time I have to contribute to my OSS projects. I'm going to close this as a duplicate, as I think searching for nvim-tree in this repo will bring up a few more previous issues where I've answered in more detail.

Issue #97, opened 2021-09-21
Is it possible to open the terminal without moving NvimTree?


davidlunadeleon:
Steps to reproduce:

1. Open file and open NvimTree.
2. Toggle terminal to open it.

Current behavior:

1. The terminal occupies the whole bottom of the screen, moving the window for NvimTree up.
2. Toggling Nvim-Tree to close and open again results in the expected layout. The tree to the side, and both terminal and file next to it.
3. If the terminal is closed and opened again, the layout returns to NvimTree not being correctly displayed again.

Expected behavior:

- Opening the terminal, doesn't affect the positioning of NvimTree if it's open.
- If NvimTree is open, the terminal doesn't take the whole bottom of the screen.
- If NvimTree is closed, the terminal can take the whole bottom of the screen.

My question is. Is this something that can already be configured? Sorry if it's possible and I missed it 😅

akinsho:
Hi @davidlunadeleon,

Not really, this is a lot more complicated than it seems for a few different reasons,
generally sidebars are just splits like any other, although it's an increasing trend in nvim to have sidebars and things, nvim doesn't treat these windows any differently from other windows, so any behaviour for splits affects all splits. Technically, we could read the window layout as a nested list and try and see if the terminal is beside a particular buffer but having done that in another plugin, I'm 100% against the complexity that brings.

Currently, the way this plugin splits windows is that it's designed to split under all existing windows (or if vertical beside all horizontal), otherwise if you had 2 vertical splits and toggled the terminal it would appear under just one of the vertical splits which is not acceptable IMO and because of the earlier point not solvable (easily/desirably).

In short, there isn't an easy solution to the problem and the only potential solution introduces a large amount of complexity which I think far outweighs the benefits. This sort of problem would be easier to deal with if nvim special cased sidebar splits, so they weren't affected by `wincmd` in the same way, but that's super unlikely to get merged upstream.

davidlunadeleon:
Thank you for the detailed answer! 🙂

It really sounds more complicated than I initially thought it would be, and definitely not something that would be solved in here. Guess that the people migrating from VSCode and other editors are used to the sidebar being there.

In case anyone wants something similar in the future, I'll leave my hacky solution in here:
```
vim.api.nvim_set_keymap('n', '<C-n>', '<cmd>:NvimTreeToggle<cr>', {})
[...]

PR #388, opened 2023-02-03
horizontal split with respect to sidebar plugin

YuCao16:
Hi,

I find the way `toggleterm` currently splits window horizontally a bit strange, it splits an area underneath all windows, which includes plug-ins like sidebar (`nvimtree`, `tagbar` etc.), but this is not how the terminal behaves in other IDE.

I have seen similar issues (#97 #359 ) and perhaps it would be worthwhile to add such an option.

Here is my solution, I can add an option to the `config.lua` to avoid breaking change if you would like to accept this PR.

### DIFF
--- a/lua/toggleterm/ui.lua
+++ b/lua/toggleterm/ui.lua
@@ -291,7 +291,11 @@ function M.open_split(size, term)
     api.nvim_set_current_win(split_win)
     vim.cmd(commands.existing)
   else
-    vim.cmd(commands.new)
+    if term.direction == "horizontal" then
+      vim.cmd("split | resize " .. config.size .. " | wincmd p")
+    else
+      vim.cmd(commands.new)
+    end
   end
 
   M.resize_split(term, size)


kaboussi (approved):

akinsho:
@YuCao16 thanks for the PR although I don't think it will work and has previously been raised. The main reason I chose the current strategy for toggleterm is because if you split per window then adding more splits to an already open toggleterm doesn't work as well because there isn't much space for new splits if opened in window that happens to be small. Also I don't want the terminals to seem to be scoped to the window. That isn't the point they are general terminals not per window terminals.

Also it isn't really reasonable to expect neovim to work like other IDEs because it isn't designed like other IDEs for example despite it being kind of annoying special filetree windows are just windows there is no reason/ no way for me to easily open the terminal but not also open it under the file tree because neovim doesn't special case these buffers.

Anyway as it stands this PR is more likely to break things and the only proper way to really "respect" sidebars would be to have either builtin respect for sidebars when splitting which I don't think the nvim team are likely to add anytime soon or adding complicated logic to check if there is some sort of specified file tree in a window by looping through all the windows then if so you split like it as it does before the switch to the file tree window and use something like `wincmd H` to re-extend it's size

YuCao16:
> @YuCao16 thanks for the PR although I don't think it will work and has previously been raised. The main reason I chose the current strategy for toggleterm is because if you split per window then adding more splits to an already open toggleterm doesn't work as well because there isn't much space for new splits if opened in window that happens to be small. Also I don't want the terminals to seem to be scoped to the window. That isn't the point they are general terminals not per window terminals.
> 
> Also it isn't really reasonable to expect neovim to work like other IDEs because it isn't designed like other IDEs for example despite it being kind of annoying special filetree windows are just windows there is no reason/ no way for me to easily open the terminal but not also open it under the file tree because neovim doesn't special case these buffers.
> 
> Anyway as it stands this PR is more likely to break things and the only proper way to really "respect" sidebars would be to have either builtin respect for sidebars when splitting which I don't think the nvim team are likely to add anytime soon or adding complicated logic to check if there is some sort of specified file tree in a window by looping through all the windows then if so you split like it as it does before the switch to the file tree window and use something like `wincmd H` to re-extend it's size

Hi @akinsho ,
Sorry for this late reply. I hadn't thought of this, and it does seem like this change was too hasty. I hope neovim can add sidebar option so that this problem can be solved once and for all.

YuCao16:
> > @YuCao16 thanks for the PR although I don't think it will work and has previously been raised. The main reason I chose the current strategy for toggleterm is because if you split per window then adding more splits to an already open toggleterm doesn't work as well because there isn't much space for new splits if opened in window that happens to be small. Also I don't want the terminals to seem to be scoped to the window. That isn't the point they are general terminals not per window terminals.
> > Also it isn't really reasonable to expect neovim to work like other IDEs because it isn't designed like other IDEs for example despite it being kind of annoying special filetree windows are just windows there is no reason/ no way for me to easily open the terminal but not also open it under the file tree because neovim doesn't special case these buffers.
> > Anyway as it stands this PR is more likely to break things and the only proper way to really "respect" sidebars would be to have either builtin respect for sidebars when splitting which I don't think the nvim team are likely to add anytime soon or adding complicated logic to check if there is some sort of specified file tree in a window by looping through all the windows then if so you split like it as it does before the switch to the file tree window and use something like `wincmd H` to re-extend it's size
> 
> So is it possible to allow users to specify which filetypes should be respected? For example, bufferline.nvim can specify the filetypes (such as "neo-tree") to have its own offsets bufferline (such as "File Explorer").

You could give `edgy.nvim` a try, even though it's not stable at this stage.

PR #396, opened 2023-02-15
doc(README): add lazy.nvim installation example

Isti115:


README.md
```diff
@@ -74,6 +74,18 @@ use {"akinsho/toggleterm.nvim", tag = '*', config = function()
...
 ```
 
+Using `lazy.nvim` in lua
+
+```lua
+require("lazy").setup({
+  {
+    "akinsho/toggleterm.nvim",
+    config = true,
+    keys = { { "<C-\\>", "<Cmd>ToggleTerm<CR>", mode = { "n", "t" } } },
```
akinsho, line 84: I don't recommend using the keys like this since the actual mapping is not the same as what you have here. I recommend just removing the keys key entirely and keeping it simple. I'd rather not suggest lazy loading as a default since this plugin already takes pains to reduce it's upfront costs

akinsho (changes requested):
Thanks for the PR, just made a comment to simplify the setup

README.md
```diff
@@ -74,6 +74,18 @@ use {"akinsho/toggleterm.nvim", tag = '*', config = function()
 end}
 ```
 
+Using `lazy.nvim` in lua
+
+```lua
+require("lazy").setup({
```
akinsho, line 80: Also I'm not sure if showing the user that they have to setup lazy is a good idea I'm not sure if a new user will assume that they need a completely separate setup call just to setup this plugin

Isti115: I was also debating this, but checked how other plugins handle it in their READMEs and decided to stick with this version after seeing the references.

README.md
Isti115: Feel free to change this as you see fit, I think that including a preset key in the copy-pasteable config greatly reduces the friction of the initial user experience, but I also think that I see what you mean by users being less likely to change it themselves if it's there by default.

akinsho: To be honest, I don't really think adding a lazy setup is especially really necessary. I think it's similar enough to packer, that people can figure it out. Since this is your contribution, it's up to you to tweak it, but as it stands, I think the keys key is not just unnecessary. I think it's actually wrong, which is what I'm trying to suggest that the key that you specified is not correct because that is not what the default mapping of toggleterm does.

PR #397, opened 2023-02-16
feat: open last toggled term on smart toggle

stevanmilic:
I find myself often wanting to open last used (toggled) terminal with the open mapping, as it feels natural to continue where you left off. Right now the first toggled terminal will be opened instead, which usually makes me re-think where I left and then I have to toggle that specific terminal that I've used.

This PR adds logic to `smart_toggle` to open last toggled terminal if none are opened, instaed of openning the first one. The changes are minimal, and I feel it doesn't complicate the code much. @akinsho  let me know what you think :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,8 +115,8 @@ I'm also going to be pretty conservative about what I add.
 
 This plugin must be explicitly enabled by using `require("toggleterm").setup{}`
 
-Setting the `open_mapping` key to use for toggling the terminal(s) will set up mappings for _normal_ mode
-If you prefix the mapping with a number that particular terminal will be opened.
+Setting the `open_mapping` key to use for toggling the terminal(s) will set up mappings for _normal_ mode.
+If you prefix the mapping with a number that particular terminal will be opened. Otherwise if a prefix is not set, then the last toggled terminal will be opened.
 
 If you set the `insert_mappings` key to true, the mapping will also take effect in insert mode; similarly setting `terminal_mappings` to will have the mappings take effect in the opened terminal.
 
--- a/doc/toggleterm.txt
+++ b/doc/toggleterm.txt
@@ -93,8 +93,9 @@ SETUP ~
 This plugin must be explicitly enabled by using `require("toggleterm").setup{}`
 
 Setting the `open_mapping` key to use for toggling the terminal(s) will set up
-mappings for _normal_ mode If you prefix the mapping with a number that
-particular terminal will be opened.
+mappings for _normal_ mode. If you prefix the mapping with a number that
+particular terminal will be opened; Otherwise if a prefix is not set, then the
+last toggled terminal will be opened.
 
 If you set the `insert_mappings` key to true, the mapping will also take effect
 in insert mode; similarly setting `terminal_mappings` to will have the mappings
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -115,8 +115,8 @@ I'm also going to be pretty conservative about what I add.
 
 This plugin must be explicitly enabled by using `require("toggleterm").setup{}`
 
-Setting the `open_mapping` key to use for toggling the terminal(s) will set up mappings for _normal_ mode
-If you prefix the mapping with a number that particular terminal will be opened.
+Setting the `open_mapping` key to use for toggling the terminal(s) will set up mappings for _normal_ mode.
+If you prefix the mapping with a number that particular terminal will be opened; in case prefix is not set, then the last toggled terminal will be opened.
```
akinsho, line 119:
```suggestion
If you prefix the mapping with a number that particular terminal will be opened. Otherwise if a prefix is not set, then the last toggled terminal will be opened.
````

File: doc/toggleterm.txt [hunk 2]
```diff
@@ -93,8 +93,9 @@ SETUP ~
 This plugin must be explicitly enabled by using `require("toggleterm").setup{}`
 
 Setting the `open_mapping` key to use for toggling the terminal(s) will set up
-mappings for _normal_ mode If you prefix the mapping with a number that
-particular terminal will be opened.
+mappings for _normal_ mode. If you prefix the mapping with a number that
+particular terminal will be opened; in case prefix is not set, then the
```
akinsho, line 97: nit: I'd update this as well similar to the other comment

akinsho (commented):
@stevanmilic thanks for the PR this looks good generally just made a small comment about the phrasing in the docs.

Also could you add a test in `tests/*_spec.lua` that just validates that if you toggle a few terminals and reopen one it does what you expect

stevanmilic:
@akinsho I made the changes. Also after writing tests I realized there's no need to save the `last_toggle_id` on `open` as it is enough to save it just on `close`, since it'll get overwritten anyway.

README.md
```diff
@@ -115,8 +115,8 @@ I'm also going to be pretty conservative about what I add.
 
 This plugin must be explicitly enabled by using `require("toggleterm").setup{}`
 
-Setting the `open_mapping` key to use for toggling the terminal(s) will set up mappings for _normal_ mode
-If you prefix the mapping with a number that particular terminal will be opened.
+Setting the `open_mapping` key to use for toggling the terminal(s) will set up mappings for _normal_ mode.
+If you prefix the mapping with a number that particular terminal will be opened. Othrwise if a prefix is not set, then the last toggled terminal will be opened.
```
akinsho, line 119:
```suggestion
If you prefix the mapping with a number that particular terminal will be opened. Otherwise if a prefix is not set, then the last toggled terminal will be opened.
```

doc/toggleterm.txt
```diff
@@ -93,8 +93,9 @@ SETUP ~
 This plugin must be explicitly enabled by using `require("toggleterm").setup{}`
 
 Setting the `open_mapping` key to use for toggling the terminal(s) will set up
-mappings for _normal_ mode If you prefix the mapping with a number that
-particular terminal will be opened.
+mappings for _normal_ mode. If you prefix the mapping with a number that
+particular terminal will be opened; Othrwise if a prefix is not set, then the
```
akinsho, line 97:
```suggestion
particular terminal will be opened; Otherwise if a prefix is not set, then the
```

akinsho:
Thanks for the PR @stevanmilic

stevanmilic:
Thank you! The codebase is really good 👌🏻

PR #399, opened 2023-02-20
feat: restore terminal view on smart toggle

stevanmilic:
A continuation of #397 where the idea is taken a bit further, and a complete terminal view is being toggled instead of just the last toggled terminal. When working with the terminals you often want to save your complete view that you've been working on, instead of just closing terminals one by one, and opening them up (again) one by one. There's already a utility mapping that let's you open and close all of your terminals, but this doesn't take into consideration what was previously opened. Therefore to make `smart_toggle` even smarter we could save the current terminal view, and then open it up the next time it is triggered.

In order to achieve this functionality we could keep a list of terminals that were closed, and use it to go over the list to open them on next smart toggle.

### DIFF
--- a/lua/toggleterm.lua
+++ b/lua/toggleterm.lua
@@ -50,24 +50,21 @@ local function setup_global_mappings()
   end
 end
 
---Create a new terminal or close beginning from the last opened
+-- Creates a new terminal if none are present or closes terminals that are
+-- currently opened, or opens terminals that were previously closed.
 ---@param _ number
 ---@param size number?
 ---@param dir string?
 ---@param direction string?
 local function smart_toggle(_, size, dir, direction)
-  local terminals = terms.get_all()
-  if not ui.find_open_windows() then
-    -- Re-open the last terminal used (if exists) or the first terminal toggled.
-    local term_id = terms.get_last_toggled_id() or terms.get_toggled_id()
-    terms.get_or_create_term(term_id, dir, direction):open(size, direction)
-  else
-    -- count backwards from the end of the list
-    for i = #terminals, 1, -1 do
-      local term = terminals[i]
-      if term and ui.term_has_open_win(term) then return term:close() end
+  local has_open, windows = ui.find_open_windows()
+  if not has_open then
+    if not ui.open_terminal_view(size, direction) then
+      local term_id = terms.get_toggled_id()
+      terms.get_or_create_term(term_id, dir, direction):open(size, direction)
     end
-    utils.notify("Couldn't find a terminal to close", "error")
+  else
+    ui.close_and_save_terminal_view(windows)
   end
 end
 
@@ -79,6 +76,8 @@ local function toggle_nth_term(num, size, dir, direction)
   local term = terms.get_or_create_term(num, dir, direction)
   ui.update_origin_window(term.window)
   term:toggle(size, direction)
+  -- Save the terminal in view if it was last closed terminal.
+  if not ui.find_open_windows() then ui.save_terminal_view({ term.id }, term.id) end
 end
 
 ---Close the last window if only a terminal *split* is open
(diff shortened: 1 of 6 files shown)


lua/toggleterm.lua
```diff
@@ -21,6 +21,10 @@ local AUGROUP = "ToggleTermCommands"
 -----------------------------------------------------------
 local M = {}
 
+---@type number[]
+-- A list of terminal IDs that are saved from the view on smart toggle.
+local terminal_view = {}
```
akinsho, line 26: Seems at this point that there should maybe be a table called `state` which holds the various bits of internal state in this file, so it's easier to see at a glance what state is being depended on centralise changing it

lua/toggleterm.lua
```diff
@@ -50,24 +54,41 @@ local function setup_global_mappings()
...
+    if not vim.tbl_isempty(terminal_view) then
+      for _, term_id in pairs(terminal_view) do
+        local term = terms.get(term_id)
+        if term then
+          term:open(size, direction)
+          opened = true
+        end
+      end
+      terminal_view = {}
+    end
```
akinsho, line 76: Could this logic be factored out into well named helper so it could easily be wrapped in an option, the thinking being if a large enough number of people push back because they liked the old mechanism or they just got used to it etc. it would be nice to be able to configure it on or off but by default my preference is to just change the behaviour without a bunch of options and complexity and see what happens

akinsho (changes requested):
@stevanmilic thanks for the PR, again generally a fan of the direction as I've definitely hit this use case of wanting to hide and restore the whole setup and is definitely in keeping with the general vision of the plugin.

I just made a comment about whether it was possible to implement the change in a way that could make it easy to configure on/off later in case a bunch of people raise pitchforks after it's merged and want some way to turn this behaviour off.

lua/toggleterm.lua
akinsho: Actually just noticed this is the toggleterm file, I wonder if the state can be held in the terminal file which is where I think of as where the internal terminal logic is done, or perhaps in the ui file which is where I think of the ui helpers and perhaps state also living in any case the main issue is having state in different places rather than centralised

stevanmilic: good point, I was thinking about this, and since it's directly tied to the `toggle` functionality I kept it in the `toggleterm` module. But I get what you mean, and I'll transfer saving/opening logic to `terminal` module and we can keep the state there as well.

lua/toggleterm.lua
stevanmilic: > Could this logic be factored out into well named helper so

sure, I'll do it in a helper.

> but by default my preference is to just change the behaviour without a bunch of options and complexity and see what happens

Exactly my thinking, as this feature seems to me as a sensible default, because it's usually want you want and how other IDEs behave in this regard. I would maybe go with the change and make it non-configurable and then if there are complaints we can make it on/off, as you wish :D

lua/toggleterm.lua
stevanmilic: done, I've moved this logic/state to `ui` module

lua/toggleterm.lua
stevanmilic: done, this is moved to `ui.open_terminal_view` function

akinsho (dismissed):
Just gave it a try locally and it works quite nicely 👍🏿

stevanmilic:
I just found an issue tho, the focus is not preserved after view is being restored with multiple terminals. It always brings focus to last opened, I'll need to keep that info in the state as well. let me push a fix.

stevanmilic:
done, should be good now. I've added `focus_term_id` in the state.

akinsho:
Alright 🚀 thanks for catching that last bug and fixing and for the contribution 🙏🏿. Merging, 🤞🏿 we don't get a lot of complaints about the change in behaviour

PR #412, opened 2023-03-12
features/add-float-titles

akinsho:
Not sure what shape this should take yet/what is actually valuable here. This will be user configurable but by default the idea is to show the terminal ID, I see floatterm shows a count and this was partially inspired by #363 by actually think this might be a bit pointless (IMO). Why is it useful to know that this is terminal 1/5 or 3/7? what do you do with that information? cc @Blackmorse you originally raised that issue and might be interested in testing this out.

### DIFF
--- a/lua/toggleterm/terminal.lua
+++ b/lua/toggleterm/terminal.lua
@@ -44,13 +44,23 @@ end
 ---@type Terminal[]
 local terminals = {}
 
+---@class FloatOpts
+---@field border string | string[]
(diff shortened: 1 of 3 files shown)


lmartinpierrat:
@akinsho is there any appetites to merge this feature ? I implemented something similar locally, but by re-using the concept of name_formatter instead.

I'm not particularly interested in the count, but just to display the terminal id somewhere else than the statusline.

Thanks !

PR #418, opened 2023-03-25
feat: support z-index option for floating windows

b0o:
When using toggleterm and [incline](https://github.com/b0o/incline.nvim) together, this change allows the user to give toggleterm a higher zindex than incline to prevent undesired overlapping.


akinsho (approved):
LGTM 👍🏿 thanks for the PR @b0o

PR #423, opened 2023-04-01
feat(config): allow `shell` parameter to be a function

jakubkaczor:
This enables determining the default shell dynamically. Please let me know if you are ok with such a change. If so, I will update the documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,7 +194,8 @@ require("toggleterm").setup{
   persist_mode = true, -- if set to true (default) the previous terminal mode will be remembered
   direction = 'vertical' | 'horizontal' | 'tab' | 'float',
   close_on_exit = true, -- close the terminal window when the process exits
-  shell = vim.o.shell, -- change the default shell
+   -- Change the default shell. Can be a string or a function returning a string
+  shell = vim.o.shell,
   auto_scroll = true, -- automatically scroll to the bottom on terminal output
   -- This field is only relevant if direction is set to 'float'
   float_opts = {
--- a/lua/toggleterm/config.lua
+++ b/lua/toggleterm/config.lua
@@ -28,7 +28,7 @@ local function shade(color, factor) return colors.shade_color(color, factor) end
 --- @field close_on_exit boolean
 --- @field direction  '"horizontal"' | '"vertical"' | '"float"'
 --- @field shading_factor number
---- @field shell string
+--- @field shell string|fun():string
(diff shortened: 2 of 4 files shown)


akinsho:
@jakubkaczor thanks for the PR. Personally I don't mind at all. If you could make this query stylistically the same as other similar options that'd be great, I'm not sure the other similar options call the function inline like that. Maybe they do, on mobile so haven't checked

jakubkaczor:
@akinsho, what do you mean by “stylistically the same”? In terms of documentation?

Also, what does it mean for an option to call a function? The only function field in the configuration I located is `winbar.name_formatter`, which is directly called. Are you talking about the type checking of the configuration field? With this change, `shell` is the only field that is of a type that is a union of other types, so I don't think similarity in the field value use could be achieved.

jakubkaczor:
Excuse me. The `size` field can also be a function, but in the code, in the comment around declaration of the field it is documented as a number, hence my mistake. Do you mean to use something similar as `ui._resolve_size` function? If so, is it needed?

akinsho:
@jakubkaczor never mind about my comment I didn't mean to cause confusion. I thought maybe this plugin declared functions in a different way I'll have a look at this again

akinsho:
@jakubkaczor never mind about my comment I didn't mean to cause confusion. I thought maybe this plugin declared functions in a different way I'll have a look at this again

akinsho (approved):
LGTM

akinsho:
@jakubkaczor can you add documentation about this feature in the help docs and README.

Would also be nice to have a small test case that essentially launches a terminal with a function `cmd` and checking that it produces the correct command.

jakubkaczor:
> @jakubkaczor can you add documentation about this feature in the help docs and README.
> 
> Would also be nice to have a small test case that essentially launches a terminal with a function `cmd` and checking that it produces the correct command.

Yes. You asked if I could add documentation to the help docs. Isn't it auto-generated by panvimdoc? Therefore, I left that out. If I shouldn't have, please clarify. 😉 And please, let me know if the test is in the form you expected.

akinsho:
@jakubkaczor changes look good thanks for making them, and the test case is fine, just wanted something in place to safeguard this feature against future changes in case myself or someone else accidentally forgets that command can be a function and it gets removed at some point

jakubkaczor:
I understand. Thanks for explanation. 😉

Issue #80, opened 2021-08-07
Feature Idea: Terminals List that can be used with picker (e.g: Telescope)


kiuKisas:
That could be great to have a list of opens terminal and to choose between them through a picker like Telescope for example.

akinsho:
This is a good idea, happy to review a PR for this 👍🏿

tknightz:
@kiuKisas Hey! I made a telescope extension, you can check it out [telescope-termfinder](https://github.com/tknightz/telescope-termfinder.nvim)

kiuKisas:
Awesome ! I'll give a try :)

PR #429, opened 2023-04-16
feat: add a terminal select command

akinsho:
This PR adds a `TerminalSelect` command which uses `vim.ui.select` to allow the user to choose which terminal to open or focus


akinsho:
fixes #80 or at least goes as far as I'm willing to go, there's also an extension for telescope so users definitely have options

PR #438, opened 2023-05-08
feat: Allow changing terminal dir in background

morgsmccauley:
The current implementation will move focus to the terminal window when changing its directory. It would be nice to be able to do that in the background.


akinsho (approved):
LGTM

PR #441, opened 2023-05-26
Add: Full mouse support for toggleterm

Zeioth:
This patch set mouse mode to "" while ToggleTerm is focused, and restores the mouse mode when focus leaves ToggleTerm. This allows the mouse to behave normally while inside ToggleTerm.

This fixes the current behavior where scrolling provokes the ToggleTerm buffer to scroll, instead of the terminal itself.

I've left the comments on this patch so it's easier to review. Please feel free to delete them or change anything you consider.

### DIFF
--- a/lua/toggleterm.lua
+++ b/lua/toggleterm.lua
@@ -332,7 +332,7 @@ local function setup_autocommands(_)
     callback = apply_colors,
   })
   
-   -- PATCH: Add mouse support for ToggleTerm
+  -- PATCH: Add mouse support for ToggleTerm
   local function set_mouse_with_toggleterm()
     local n_mouse = vim.o.mouse
 


lua/toggleterm.lua
```diff
@@ -352,22 +352,23 @@ local function setup_autocommands(_)
...
-        callback = function() vim.fn.system "tmux set mouse off" end,
-      })
-
-      -- Enable tmux mouse when mouse leaves toggleterm
-      api.nvim_create_autocmd({ "WinLeave <buffer>" }, {
-        desc = "Enable tmux mouse when mouse leaves toggleterm",
-        group = AUGROUP,
-        callback = function() vim.fn.system "tmux set mouse on" end,
-      })
+    if os.getenv "TMUX" ~= nil then
```
akinsho, line 355:
```suggestion
    if vim.env.TMUX then
```
Should be enough for this

lua/toggleterm.lua
```diff
@@ -352,22 +352,23 @@ local function setup_autocommands(_)
...
+      local output = vim.fn.system 'tmux display -p "#{mouse}"'
```
akinsho, line 356: Would be good to have a comment on what this command is doing? Also if it's slow it might be better to use `vim.fn.jobstart`

lua/toggleterm.lua
```diff
@@ -352,22 +352,23 @@ local function setup_autocommands(_)
...
+      if output:sub(1, 1) == "1" then
+        -- Disable tmux mouse while using toggleterm
+        autocmd({ "TermEnter", "WinEnter <buffer>" }, {
```
akinsho, line 359: If an autocommand is buffer local it should pass in the buffer as an argument to the function not using `<buffer>`. I think this should be a separate autocmd with `buffer = num`

akinsho (changes requested):
@Zeioth thanks for the PR haven't a huge amount of time to test it out but made some comments re. implementation. Also if we are going to mess with user settings for external programs like this then this should be gated by a configuration option although I'm happy to default it to true.

lua/toggleterm.lua
Zeioth: that command prints 1 if mouse support is enabled in tmux. We are saying "If we are running tmux, and mouse support is enabled, then create this autocmds".

That's why we need it to be blocking. It is fast though, and it doesn't cause issues if the result is not the expected.

This patch is already been [merged in rnvimr](https://github.com/kevinhwang91/rnvimr/pull/150) in case you wanna check out. I just changed the names to match toggleterm.

lua/toggleterm.lua
Zeioth: Running the same code without "WinEnter <buffer>" achieves the same, but only after moving the mouse at least once. So the event seems to be working as it should.

I've tried removing the <buffer> part as [it does not appear in the documentation](https://neovim.io/doc/user/autocmd.html#WinEnter) and still seems to work as expected. So it is probably ignored by neovim for better backwards compatibility. I'm gonna leave it as just "WinEnter".

Also, you probablly saw it but the most important part of the code is 

``` lua
if vim.bo.filetype == "toggleterm" then
```
Meaning wathever we do is only gonna affect toggleterm.

Zeioth:
Please test it when you have time. All should be fine now. I've added the config option

    mouse_support = true

In my machine everything seems fine both on tmux and my normal terminal.

lua/toggleterm.lua
```diff
@@ -331,6 +331,60 @@ local function setup_autocommands(_)
     pattern = "term://*",
     callback = apply_colors,
   })
+  
+  -- PATCH: Add mouse support for ToggleTerm
+  local function set_mouse_with_toggleterm()
+    local n_mouse = vim.o.mouse
+
+    -- Disable nvim mouse support while we are on the toggleterm buffer
+    if string.match(n_mouse, "[a|h|n]") then
```
akinsho, line 340:
```suggestion
    if n_mouse:match("[a|h|n]") then
```

akinsho (commented):
@Zeioth I only just realised this but appears all of this logic is just to allow you to disable the mouse behaviour inside rnvimr. In which case I'm sorry but I won't actually be accepting this PR. This plugin must never contain custom logic for specific terminal applications it should server as an abstraction for the terminal that can be used to use different programs but I absolutely do not want to maintain custom logic targeted at specific programs this is completely against what I view this plugin as being for.

In your case why not just create a specific terminal for rnvimr and then disable the mouse during that terminals on open and re-enable it on close or something. This doesn't have to be built into this plugin

akinsho:
@Zeioth please also look at `vim.o.mousescroll` you can disable what you want in there as you choose in either case this PR I think is not necessary. Sorry been too busy to look properly or I would have pointed this out earlier

akinsho:
@Zeioth there is a wiki for this plugin for exactly this type of thing please have a look in https://github.com/akinsho/toggleterm.nvim/wiki you should be able to create new pages and edit existing information

Zeioth:
I will do. Thank you so much @akinsho

PR #449, opened 2023-06-23
fix: replace vim.wo with nvim_set_option_value

AThePeanut4:
I recently discovered [stickybuf.nvim](https://github.com/stevearc/stickybuf.nvim), and it's great at preventing buffers from opening inside plugin windows, ToggleTerm included.

However, due to the way ToggleTerm sets its window options, if I accidentally/intentionally open a new file while in the ToggleTerm buffer, the new buffer inherits ToggleTerm's window options (specifically `winhighlight`, `nonumber` and `norelativenumber` for my config). This means that when stickybuf moves that buffer to a different window, it keeps those window options and thus has a dark background and no line numbers. The same kind of thing also happens without stickybuf, but opening buffers inside ToggleTerm's window causes some other glitchy behaviour if you immediately then try and open ToggleTerm using the keybinding again, which is sort of to be expected.

The root cause is the use of `vim.wo` to set window options, which behaves counter-intuitively (in my opinion): see https://github.com/folke/lazy.nvim/issues/829 and https://github.com/neovim/neovim/issues/14595. For window-local options, `vim.wo` sets both the global and local values of the option, like `:set` does. This is different to what I would have expected, which is what `vim.bo` does - it only sets the local value for buffer-local options, like `:setlocal` does. For global-local options, `vim.wo` and `vim.bo` both do the same and act like `:setlocal`, only setting the local value of the option.

This therefore replaces all uses of `vim.wo` to set window-local options with `nvim_set_option_value`, so that only the local value is set, meaning that new buffers created in the ToggleTerm window will not inherit those options. I haven't changed the lines where `winbar` or `sidescrolloff` are set, since those are global-local options and thus `vim.wo` works as expected. Those can also easily be changed for consistency if needed.

I also haven't changed any uses of `vim.wo` to get option values (in tests), since that also works as expected as far as I can tell.

### DIFF
--- a/lua/toggleterm/terminal.lua
+++ b/lua/toggleterm/terminal.lua
@@ -415,15 +415,14 @@ end
 
 ---@package
 function Terminal:__set_win_options()
-  local win = vim.wo[self.window]
   if self:is_split() then
     local field = self.direction == "vertical" and "winfixwidth" or "winfixheight"
-    win[field] = true
+    vim.api.nvim_set_option_value(field, true, { scope = "local", win = self.window })
   end
 
   if config.hide_numbers then
-    win.number = false
-    win.relativenumber = false
+    vim.api.nvim_set_option_value("number", false, { scope = "local", win = self.window })
+    vim.api.nvim_set_option_value("relativenumber", false, { scope = "local", win = self.window })
   end
 end
 
--- a/lua/toggleterm/terminal.lua
+++ b/lua/toggleterm/terminal.lua
@@ -417,12 +417,12 @@ end
 function Terminal:__set_win_options()
   if self:is_split() then
     local field = self.direction == "vertical" and "winfixwidth" or "winfixheight"
-    vim.api.nvim_set_option_value(field, true, { scope = "local", win = self.window })
+    utils.wo_setlocal(self.window, field, true)
   end
 
   if config.hide_numbers then
-    vim.api.nvim_set_option_value("number", false, { scope = "local", win = self.window })
-    vim.api.nvim_set_option_value("relativenumber", false, { scope = "local", win = self.window })
+    utils.wo_setlocal(self.window, "number", false)
+    utils.wo_setlocal(self.window, "relativenumber", false)
   end
 end
 
--- a/lua/toggleterm/ui.lua
+++ b/lua/toggleterm/ui.lua
@@ -139,7 +139,11 @@ function M.hl_term(term)
     return hi_target
   end, hl_names)
 
-  vim.wo[window].winhighlight = table.concat(highlights, ",")
+  vim.api.nvim_set_option_value(
+    "winhighlight",
+    table.concat(highlights, ","),
+    { scope = "local", win = window }
+  )
 end
 
 ---Create a terminal buffer with the correct buffer/window options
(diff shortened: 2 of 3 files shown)


akinsho:
Hi @AThePeanut4 thanks for the PR and the detailed explanation. I've actually been aware of this issue but initially the `option_value` helpers didn't exist then they did and under the hood, `vim.[w|b]o` use them actually but there is still https://github.com/neovim/neovim/pull/20288 which I've actually been waiting for since I actually find the method incredibly verbose and the bug has existed for quite a long time before this.

Whilst generally a fix is desirable I'm a little reticent to have this overhead of knowing it's safe to use `vim.bo` in some places but otherwise using a very longwinded `set_option_value` in others.

Which is not to say this isn't useful that I'm open to merging just something I want to think over since to be honest my ideal solution would just be to wait for that PR to get merged then change `vim.wo[1][0].winhighlights` instead etc.

akinsho (changes requested):
@AThePeanut4 given the verbosity of the `set_option_value` function could you create a helper in the `utils` module called something like `setlocal` and then add a TODO to that function that it should be replaced eventually by double indexing (hopefully whenever that is merged)

AThePeanut4:
> @AThePeanut4 given the verbosity of the `set_option_value` function could you create a helper in the `utils` module called something like `setlocal` and then add a TODO to that function that it should be replaced eventually by double indexing (hopefully whenever that is merged)

Done, I called it `wo_setlocal`.

> Hi @AThePeanut4 thanks for the PR and the detailed explanation. I've actually been aware of this issue but initially the `option_value` helpers didn't exist then they did and under the hood, `vim.[w|b]o` use them actually but there is still [neovim/neovim#20288](https://github.com/neovim/neovim/pull/20288) which I've actually been waiting for since I actually find the method incredibly verbose and the bug has existed for quite a long time before this.

`nvim_set_option_value` is rather verbose yeah, but until that PR merges it seems to be the only Lua way to `:setlocal` window-local options. I hadn't seen that PR before, it would definitely allow for a nicer solution than what I'm doing here.

I'm fairly new to Vim/Neovim so I spent quite a long while figuring out exactly how all the different kinds of options work and what each way of changing them actually does. While doing this I did indeed find the runtime `_meta.lua` file where these `vim.*` interfaces are defined in terms of `nvim_set_option_value`.

> Whilst generally a fix is desirable I'm a little reticent to have this overhead of knowing it's safe to use vim.bo in some places but otherwise using a very longwinded set_option_value in others.

I assume you're referring to `vim.wo` here, `vim.bo` should always be safe to use, in the sense that it always sets the local value only. I'm not sure what you mean by overhead - using `nvim_set_option_value` calls the C API directly rather than through `vim.wo`. Unless you mean overhead as in verbosity, in which case `wo_setlocal` fixes that I think.

akinsho:
Thanks for making the changes @AThePeanut4 yep by overhead I meant verbosity/wanting to just use `bo` and `wo` for all option setting for windows and buffers without having to think about the fact that if trying to use `wo` it wouldn't work without "bleeding". Anyway this solves this for now and hopefully I can refactor it once that PR is merged and stable.

Issue #453, opened 2023-07-03
[BUG] Vim(bnext): E85: There is no listed buffer


towry:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

```
E5108: Error executing lua vim/_editor.lua:402: nvim_exec2()..BufEnter Autocommands for "term://*#toggleterm#*": Vim(append):Error executing lua callback: vim/_editor.lua:402: nvim_exec2().
.BufEnter Autocommands for "term://*#toggleterm#*"..script nvim_exec2() called at BufEnter Autocommands for "term://*#toggleterm#*":0: Vim(bnext):E85: There is no listed buffer             
stack traceback:                                                                                                                                                                             
        [C]: in function 'nvim_exec2'                                                                                                                                                        
        vim/_editor.lua:402: in function 'cmd'                                                                                                                                               
        ...l/share/nvim/lazy/nvim-toggleterm.lua/lua/toggleterm.lua:89: in function 'close_last_window'                                                                                      
        ...l/share/nvim/lazy/nvim-toggleterm.lua/lua/toggleterm.lua:98: in function <...l/share/nvim/lazy/nvim-toggleterm.lua/lua/toggleterm.lua:95>                                         
        [C]: in function 'nvim_exec2'                                                                                                                                                        
        vim/_editor.lua:402: in function 'cmd'                                                                                                                                               
        ....local/share/nvim/lazy/mini.starter/lua/mini/starter.lua:1476: in function 'eval_fun_or_string'                                                                                   
        ....local/share/nvim/lazy/mini.starter/lua/mini/starter.lua:868: in function 'eval_current_item' 
```

### Expected Behavior

Should ignore error from `vim.cmd('bnext')`.

simillar issue: https://github.com/echasnovski/mini.nvim/issues/393

### Steps To Reproduce

This is a rare case, after the mini.starter dashboard exit with command `q`, the `mini.starter` will run `vim.cmd('bprevious')`, which will bring up the toggleterm window, then toggleterm's `BufEnter` triggered which will try to close last window(the mini.starter window)...

the root cause is this line: https://github.com/akinsho/toggleterm.nvim/blob/4cac5c96445df3a97f162a0f810e38fde1eb525e/lua/toggleterm.lua#L89

### Environment

```Markdown
- OS: macos
- neovim version: NVIM v0.10.0-dev-1904+g3bf887f6e-Homebrew
- Shell: zsh
```

PR #454, opened 2023-07-03
fix: #453

towry:


akinsho (approved):
LGTM, haven't tested it but seems like it should work. Will merge but if it goes sideways might have to revert it later

PR #456, opened 2023-07-05
Note custom terminals with no cmd option

jackfranklin:
I spent a long time trying to understand how to do this because when I read the docs I though the `cmd` argument was required. Perhaps this is me being silly, so feel free to ignore if so, but I wanted to make the README explicit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -401,6 +401,8 @@ Terminal:new {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


akinsho (approved):
LGTM 👍🏾 thanks

PR #461, opened 2023-07-10
refactor(terminal): re-use spawn function

akinsho:
Refactors the `open` function to re-use spawn rather than essentially re-implement it.


akinsho:
@jedrzejboczar as the initial author of the `spawn` functionality, can you see if this change breaks anything for you?

jedrzejboczar:
@akinsho Everything seems to work fine for me with this change.

Issue #466, opened 2023-07-31
[BUG] TermExec doesn't work with config.shell = function


asror1:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I when trying to use `TermExec cmd="some command"` I am faced with the following error: 
![image](https://github.com/akinsho/toggleterm.nvim/assets/92228287/9c367c81-3fe4-44ca-a57c-b9e34359cbf2)
And what ends up happening is the command is **not executed**. 

### Expected Behavior

When using config.shell = function, the `TermExec` should not throw an error, and instead execute the passed in `cmd` as normal. 

### Steps To Reproduce

1. In any buffer
2. With the following minimal config: 
```lua
require("toggleterm").setup({
  shell = function()
    return vim.o.shell
  end,
})
```
3. Run the following command `:TermExec cmd="echo test"`
4. And you should see the following error... 
![image](https://github.com/akinsho/toggleterm.nvim/assets/92228287/c3b9f40b-7eff-47fd-a9f8-8ebdc08164f6)

### Environment

```Markdown
- OS: Windows 10
- neovim version: 
NVIM v0.8.2
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3        
Compiled by runneradmin@fv-az28-353
- Shell: pwsh
```

### Anything else?

I believe this is an easy fix. If I'm not mistaken it comes from using a function for the **shell** option in the config. I will be submitting a fix PR for this for review.

akinsho:
fixed by #467

PR #467, opened 2023-07-31
fix: TermExec cmd with config.shell = function

asror1:
fixes(#466)

This minor change personally resolves the issue I was facing, although potentially there might be other instances in the code where the shell option is assumed to be a string, I have not looked through the codebase for other occurrences though.

### DIFF
--- a/lua/toggleterm/terminal.lua
+++ b/lua/toggleterm/terminal.lua
@@ -33,6 +33,7 @@ local function get_comment_sep() return is_windows and is_cmd(vim.o.shell) and "
 
 local function get_newline_chr()
   local shell = config.get("shell")
+  if type(shell) == "function" then shell = shell() end
   return is_windows and (is_pwsh(shell) and "\r" or "\r\n") or "\n"
 end
 


akinsho (changes requested):
@asror1 thanks for the PR this looks good, but as you say might miss places where else shell is expected to be a string. I think it would be better to re-assign the value inside of the `new` or the `spawn` function similar to how the `height` and `width` are derived and stored from functions so that the function is only evaluated once on creation

asror1:
> it would be better to re-assign the value inside of the `new` or the `spawn` function similar to how the `height` and `width` are derived and stored from functions so that the function is only evaluated once on creation

I've looked through the codebase and the `spawn` function, or rather the inner `__spawn` function is already evaluating the shell function before using it. 
```lua
function Terminal:__spawn()
  local cmd = self.cmd or config.get("shell")
  if type(cmd) == "function" then cmd = cmd() end
  ...
```
I believe the error was truly just coming the local `get_newline_chr()` function in the `terminal` module that is using the shell option directly from config, rather than having it passed in as a parameter. But I'm not sure if it would make sense to have `get_newline_chr()` accept the shell as parameter,  since it is only ever used in the `with_cr` local func, which in turn is only ever used in `Terminal:send` method.

akinsho (approved):
Will merge as is for now since this fixes the issue for now

PR #480, opened 2023-08-21
[feature] cmd=% work with spaced paths

m10090:
this should allow cmd =% use spaced paths
(sorry for any imperfection this is my frist contribute)


lua/toggleterm/commandline.lua
```diff
@@ -93,8 +93,13 @@ local term_exec_options = {
 
     for _, path in ipairs(paths) do
       -- solve space in file names
-      local glob_str = path.gsub(" ", "\\ ")
+      if path != ''
```
akinsho, line 96: This isn't valid `lua` you probably meant to write `~=`

.DS_Store
akinsho, line 1: This shouldn't be commited

lua/.DS_Store
akinsho, line 1: This is also not necessary

akinsho (changes requested):
Thanks for the PR @m10090 would be good to have a clear description in the README of what this PR fixes but I think it looks ok as a change but some things need to be tidied up

lua/toggleterm/commandline.lua
m10090: Sorry about the inconvenience and I am happy to be a part of your project

PR #482, opened 2023-08-21
fix spaced paths

m10090:
I think it is done now


m10090:
It ran all the tests and the stylua except push changes (which I think It because of the fetch)

.github/workflows/ci.yml
akinsho, line 1: These changes shouldn't be here

.DS_Store
akinsho, line 1: I think you need to add these files to your global gitignore, but in either case they shouldn't be here also

.github/.DS_Store
akinsho, line 1: Same as above

akinsho (changes requested):
Think there are a few unrelated changes here that need to be removed

m10090:
I made my `secerts.WORKFLOW_ACCESS_TOKEN` all the test got accepted except for ('doc') push change error which I think is related to repo being forked

akinsho (changes requested):
@m10090 there still appears to be several changes that aren't related to what you were trying to accomplish at all. None of the DS store files should be committed. Also. You shouldn't need to change the gitignore to add workflow files to it, the DS_store files should be in your global gitignore not here in this project. As far as I can tell the only change should be in the lua file

akinsho:
Please don't worry about the workflow because it is broken due to unrelated issues which are for me to fix. You don't need to commit any changes for that.

m10090:
yah I am sorry about that

m10090:
I will clear all what I have did recently and start a new fork with just the lua file change thank you 😊

PR #486, opened 2023-09-06
feat: add `Terminal.find` function

willothy:
Very generic way of finding a terminal that should be adaptable to most use cases. I'm using it for finding a terminal by bufnr.

### DIFF
--- a/lua/toggleterm/terminal.lua
+++ b/lua/toggleterm/terminal.lua
@@ -536,6 +536,20 @@ function M.get(id, include_hidden)
   return (term and (include_hidden == true or not term.hidden)) and term or nil
 end
 
+---Get the first terminal that matches a predicate
+---@param predicate fun(term: Terminal): boolean
+---@return Terminal?
+function M.find(predicate)
+  if type(predicate) ~= "function" then
+    utils.notify("terminal.find expects a function, got " .. type(predicate), "error")
+    return
+  end
+  for _, term in pairs(terminals) do
+    if predicate(term) then return term end
+  end
+  return nil
+end
+
 ---Return the potentially non contiguous map of terminals as a sorted array
 ---@param include_hidden boolean? whether or nor to filter out hidden
 ---@return Terminal[]


lua/toggleterm/terminal.lua
```diff
@@ -536,6 +536,16 @@ function M.get(id, include_hidden)
   return (term and (include_hidden == true or not term.hidden)) and term or nil
 end
 
+---Get the first terminal that matches a predicate
+---@param predicate fun(term: Terminal): boolean
+---@return Terminal?
+function M.find(predicate)
+  for _, term in pairs(terminals) do
+    if predicate(term) then return term end
```
akinsho, line 544: I wonder if only a well defined subset of the terminal's fields be passed ot the predicate so people can't do undesirable things or use private fields or trigger side effects here that they shouldn't 🤔. I'm sure some some people are already doing stuff like this but I worry about users getting used to operating on the whole terminal in a way that a lot of people become dependent on private fields.

The cat might already be out of the bag on that one though since lua doesn't really support privacy for fields

willothy: Yeah, I'm definitely doing a lot of janky things in my config taking advantage of Lua's lack of privacy haha (haven't needed to do so with toggleterm, though). I'm not really sure about this specific case - the idea makes sense, but the function will return a full Terminal object that will be exposed either way, so that would really only avoid mutation in the predicate.

willothy: Maybe it would be better to expose one or two functions for filtering terminals by bufnr or name, etc. and avoid using a higher order function entirely?

akinsho: 🤔 I guess it's always been the case that users can interact with the terminal directly and I have advised in the README not to use the private methods but I'm sure some users will anyway 🤦🏾‍♂️, I guess more specific functions will just lead to future requests for more variants so probably one function that is super generic is better

willothy: Alright, that makes sense. Let me know if there are any changes you'd like! 

I think with Lua it's almost impossible to make something private, even locals can be retrieved with `debug.getupvalue`, so I'm not sure it's worth spending much energy on haha

willothy:
Anything else you want changed here? If you're not interested in merging this that's alright, pretty much the same behavior can be achieved with `vim.iter(terminal.get_all()):find(...)` as long as #487 gets fixed. Happy to make any changes or close if this isn't wanted :)

akinsho:
Nope no changes needed, I'm just focused on personal projects of mine in the free time so can only afford a few mins here and there to look at neovim OSS stuff. This generally looks fine just want to make sure I don't carelessly merge (which I've done in the past) without thinking if this might produce future issues (don't think it will though)

willothy:
> Nope no changes needed, I'm just more focused on personal projects of mine in my free time so can only afford a few mins here and there to look at neovim OSS stuff. This generally looks fine just want to make sure I don't carelessly merge (which I've done in the past) without thinking if this might produce future issues (don't think it will though)

Totally, no rush!

PR #488, opened 2023-09-06
fix: determine custom terminal ids on spawn

### LINKED ISSUE #487: [BUG] Terminal id collisions with custom terminals

### Current Behavior

When creating multiple custom terminals with `Terminal:new()`, ids are assigned automatically if they're not provided.

However, the `next_id` function only checks *existing* terminals, not ones that have been newly created (not yet spawned).

This leads to terminals having identical IDs, and therefore toggleterm losing track of one of them when the other is opened.

Both will remain open, but the previous is removed from the `terminals` dict and therefore is inaccessible to API functions such as `get_all`, `get`, etc.

### Expected Behavior

Created terminals should never overwrite each other; Id conflicts should be handled properly.

### Steps To Reproduce

```lua
local term = require("toggleterm.terminal")
local Terminal = term.Terminal

local one = Terminal:new()
local two = Terminal:new()
one:open()
two:open()

vim.print(#term.get_all(true)) -- should return 2, but returns 1
```

### Environment

```Markdown
- OS: Linux
- neovim version: 0.10.0
- Shell: zsh
```

### Anything else?

Happy to contribute a PR for this, but I think it needs some discussion first. I know that the way the `next_id` function works is done that way to keep id numbers low, for ease of use. This is good and should be kept, something like a counter that increments for each new terminal is out of the question. 

<strike>So, my idea is using an additional cache set, like `terminals`, but for `reserved` terminals that have been initialized but not spawned. This set would be checked anywhere `terminals` is, to ensure that id collisions never occur. 

Another option could be only assigning terminal ids to custom terminals when they're spawned.
</strike>

See #488, assigning terminal ids on spawn was a very simple change and does not affect terminals created with a specific id.

Would appreciate some input :)

willothy:
This PR sets IDs at spawn-time for terminals that were created without an id/count specifies. See linked issue for more info.

closes #487

### DIFF
--- a/lua/toggleterm/terminal.lua
+++ b/lua/toggleterm/terminal.lua
@@ -194,7 +194,7 @@ function Terminal:new(term)
   local conf = config.get()
   self.__index = self
   term.direction = term.direction or conf.direction
-  term.id = id or next_id()
+  term.id = id
   term.float_opts = vim.tbl_deep_extend("keep", term.float_opts or {}, conf.float_opts)
   term.clear_env = term.clear_env
   term.auto_scroll = vim.F.if_nil(term.auto_scroll, conf.auto_scroll)
@@ -452,6 +452,7 @@ end
 ---Spawn terminal background job in a buffer without a window
 function Terminal:spawn()
   if not self.bufnr or not api.nvim_buf_is_valid(self.bufnr) then self.bufnr = ui.create_buf() end
+  if not self.id then self.id = self.id or next_id() end
   self:__add()
   if api.nvim_get_current_buf() ~= self.bufnr then
     api.nvim_buf_call(self.bufnr, function() self:__spawn() end)
--- a/lua/toggleterm/terminal.lua
+++ b/lua/toggleterm/terminal.lua
@@ -452,7 +452,7 @@ end
 ---Spawn terminal background job in a buffer without a window
 function Terminal:spawn()
   if not self.bufnr or not api.nvim_buf_is_valid(self.bufnr) then self.bufnr = ui.create_buf() end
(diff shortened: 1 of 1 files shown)


lua/toggleterm/terminal.lua
```diff
@@ -452,6 +452,7 @@ end
 ---Spawn terminal background job in a buffer without a window
 function Terminal:spawn()
   if not self.bufnr or not api.nvim_buf_is_valid(self.bufnr) then self.bufnr = ui.create_buf() end
+  if not self.id then self.id = self.id or next_id() end
```
akinsho, line 455: is the `self.id or next_id()` necessary in this clause? could it just be
```suggestion
  if not self.id then self.id = next_id() end
```
seems the `not self.id` has already proven that the id is falsy so no need for the or

akinsho (commented):
Small change to the logic otherwise this LGTM, thanks 👍🏾
I don't recall off the top of my head but presumably spawn is always called right so we aren't going to miss giving some terminal an ID now?

lua/toggleterm/terminal.lua
willothy: True, haha not sure how I thought that was needed

willothy:
> Small change to the logic otherwise this LGTM, thanks 👍🏾 I don't recall off the top of my head but presumably spawn is always called right so we aren't going to miss giving some terminal an ID now?

Yes, exactly. Usually spawn would be called right away, but for cases where it isn't IDs will now be assigned "lazily" so that they don't conflict. Any terminal with a predefined ID will of course have the same behavior as before.

willothy:
Any ideas for a better way of doing this since it caused some issues?

PR #497, opened 2023-09-27
fix(#496): call ToggleTermSetName with count

### LINKED ISSUE #496: [BUG] ToggleTermSetName does not work with count

### Current Behavior

Calling `ToggleTermSetName` with a count opens the dialogue to enter a new name, and typing a name and hitting `<CR>` results in the error "attempt to index local 'term' (a nil value)". 

### Expected Behavior

The specified terminal should be renamed without error.

### Steps To Reproduce

1. Minimal ```init.lua``` to reproduce the above behaviour:
```lua
local lazypath = vim.fn.stdpath('data') .. '/lazy/lazy.nvim'
if not vim.loop.fs_stat(lazypath) then
  vim.fn.system({
    'git',
    'clone',
    '--filter=blob:none',
    'https://github.com/folke/lazy.nvim.git',
    '--branch=stable', -- latest stable release
    lazypath,
  })
end
vim.opt.rtp:prepend(lazypath)

-- Lazy load plugins
require("lazy").setup {
  {
    'akinsho/toggleterm.nvim',
    version = "*",
    lazy = true,
    cmd = {
      'ToggleTerm',
      'TermSelect',
      "ToggleTermSendCurrentLine",
      "ToggleTermSendVisualLines",
      "ToggleTermSendVisualSelection",
      "ToggleTermSetName",
    },
    keys = { [[<C-`>]] },
    config = function ()
      require("toggleterm").setup {
        open_mapping = [[<C-`>]],
        insert_mappings = true, -- whether or not the open mapping applies in insert mode
        terminal_mappings = true, -- whether or not the open mapping applies in the opened terminals
        close_on_exit = true, -- close the terminal window when the process exits
        -- Change the default shell. Can be a string or a function returning a string
        shell = vim.o.shell,
        auto_scroll = true, -- automatically scroll to the bottom on terminal output
        -- This field is only relevant if direction is set to 'float'
      }
	  vim.keymap.set('n', '<A-r>', '<CMD>ToggleTermSetName<CR>', {noremap = true})
    end
  }
}
```

2. Run:
a. Create two terminals with ```<C-`><Esc>2<C-`><Esc>```.
c. Type ```2<A-r>```.

### Environment

```Markdown
- OS: Ubuntu 22.04.2 LTS running as a WSL2 distro in Windows 11
- neovim version: 0.9.2
- Shell: bash
- Terminal: wezterm
```

### Anything else?

Thanks for your work on this plugin, I rely on it quite extensively!

abhinavnatarajan:
Fixes #496.

### DIFF
--- a/lua/toggleterm.lua
+++ b/lua/toggleterm.lua
@@ -436,7 +436,9 @@ local function setup_commands()
     if no_count and no_name then
       get_subject_terminal(request_term_name)
     elseif no_name then
-      request_term_name()
+      local term = terms.get(opts.count)
+      if not term then return end
+      request_term_name(term)
     elseif no_count then
       get_subject_terminal(function(t) set_term_name(opts.args, t) end)
     else
@@ -444,7 +446,7 @@ local function setup_commands()
       if not term then return end
       set_term_name(opts.args, term)
     end
-  end, { nargs = "?" })
+  end, { nargs = "?", count = true })
 end
 
 function M.setup(user_prefs)


akinsho (approved):
LGTM 👍🏾

PR #504, opened 2023-10-28
feat: display name on floating window title

guill:
This PR adds two new options for floating windows: `title` and `title_pos`. If `title` is true, the floating window will display the terminal's name (or the ID if it has no name).  The position of the title can be controlled via `title_pos` (just like in `nvim_open_win`).


akinsho (changes requested):
LGTM but would be good to see this
1. Gated to versions of neovim that have this option available
2. Added to the documentation so it's discoverable by more users

akinsho:
fixed by #534

PR #513, opened 2023-11-22
clear terminal on windows

haeuslerd:
functionality added to clear an already opened terminal in windows.
clear function is not valid on windows os. Therefore added additional check for os.

### DIFF
--- a/lua/toggleterm/terminal.lua
+++ b/lua/toggleterm/terminal.lua
@@ -323,14 +323,21 @@ function Terminal:send(cmd, go_back)
   end
 end
 
-function Terminal:clear() self:send("clear") end
+-- Clear the terminal
+function Terminal:clear() 
+	if vim.fn.has('win32') == 1 then
+		self:send("cls")
+	else
+		self:send("clear")
+	end
+end
 
 ---Update the directory of an already opened terminal
 ---@param dir string
 function Terminal:change_dir(dir, go_back)
   dir = _get_dir(dir)
   if self.dir == dir then return end
-  self:send({ fmt("cd %s", dir), "clear" }, go_back)
+  self:send({ fmt("cd %s", dir), self.clear() }, go_back)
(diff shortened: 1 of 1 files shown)


lua/toggleterm/terminal.lua
```diff
@@ -323,14 +323,21 @@ function Terminal:send(cmd, go_back)
   end
 end
 
-function Terminal:clear() self:send("clear") end
+-- Clear the terminal
+function Terminal:clear() 
+	if vim.fn.has('win32') == 1 then
```
akinsho, line 328: Three is already a variable in this file called `is_windows` that does this check so we don't need to do this logic again.

Also I think we could just change the command e.g.
```lua
local clear = is_windows and "cls" or "clear"
self:send(clear)
```

lua/toggleterm/terminal.lua
```diff
@@ -323,14 +323,21 @@ function Terminal:send(cmd, go_back)
...
+	end
+end
 
 ---Update the directory of an already opened terminal
 ---@param dir string
 function Terminal:change_dir(dir, go_back)
   dir = _get_dir(dir)
   if self.dir == dir then return end
-  self:send({ fmt("cd %s", dir), "clear" }, go_back)
+  self:send({ fmt("cd %s", dir), self.clear() }, go_back)
```
akinsho, line 340: Generally in lua when using self methods you should use the `self:` vs `self.` syntax. If you use `self.blah` the `self` object won't be passed in/usable from the method. This can lead to super annoying/hard to track bugs

akinsho (changes requested):
Thanks for taking the time to find a solution and raise a PR. I've made a few small comments but otherwise LGTM

haeuslerd:
Thanks for the feedback. I updated the function as requested.

akinsho (approved):
LGTM 👍🏾

PR #518, opened 2023-12-04
newline_chr is optionally user-specified

fisher-j:
R on cmd.exe on windows adds an extra newline using the current logic (newline is chosen as: "\r\n"). This proposal makes `newline_chr` a property of the `Terminal`, and allows the user to specify it's value when creating custom terminals.

### DIFF
--- a/doc/toggleterm.txt
+++ b/doc/toggleterm.txt
@@ -477,13 +477,18 @@ using the current count e.g. `:5ToggleTerm<CR>`
     local lazygit = Terminal:new({ cmd = "lazygit", count = 5 })
 <
 
-You can also set a custom layout for a terminal.
+You can also set several options for the custom terminal including layout,
+opening and closing functions, and the newline charcter used when sending text
+to the terminal. The `newline_chr` defaults to "\r" on powershell, "\r\n" on
+cmd.exe, and "\n" otherwise.
 
 >lua
     local lazygit = Terminal:new({
       cmd = "lazygit",
       dir = "git_dir",
       direction = "float",
+      -- Try changing if you get too many/few blank lines when sending to terminal
+      newline_chr = "\n" 
       float_opts = {
         border = "double",
       },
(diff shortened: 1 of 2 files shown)


lua/toggleterm/terminal.lua
```diff
@@ -287,10 +290,11 @@ end
 
 ---Combine arguments into strings separated by new lines
 ---@vararg string
+---@param newline_chr string
 ---@return string
-local function with_cr(...)
+local function with_cr(newline_chr, ...)
   local result = {}
-  local newline_chr = get_newline_chr()
+  -- local newline_chr = get_newline_chr()
```
akinsho, line 297: should this be commented out? or removed?

fisher-j: I think this can safely be removed.

doc/toggleterm.txt
```diff
@@ -477,13 +477,18 @@ using the current count e.g. `:5ToggleTerm<CR>`
     local lazygit = Terminal:new({ cmd = "lazygit", count = 5 })
 <
 
-You can also set a custom layout for a terminal.
+You can also set several options for the custom terminal including layout,
+opening and closing functions, and the newline charcter used when sending text
```
akinsho, line 481:
```suggestion
opening and closing functions, and the newline character used when sending text
```

doc/toggleterm.txt
```diff
@@ -477,13 +477,18 @@ using the current count e.g. `:5ToggleTerm<CR>`
...
+to the terminal. The `newline_chr` defaults to "\r" on powershell, "\r\n" on
+cmd.exe, and "\n" otherwise.
 
 >lua
     local lazygit = Terminal:new({
       cmd = "lazygit",
       dir = "git_dir",
       direction = "float",
+      -- Try changing if you get too many/few blank lines when sending to terminal
```
akinsho, line 490: I don't think we need a special addition for this especially given the explainer above

akinsho (commented):
LGTM after the last couple of tweaks

doc/toggleterm.txt
fisher-j: removed

akinsho (approved):
LGTM, here's hoping nothing breaks

PR #532, opened 2024-01-12
doc(float_opts): add number/function description of row and col

penryu:
**UPDATE:**

A doc-only change to describe that `row` and `col` are also configurable via number or function, like `width` and `height`.

-----

**Original description:**

~This one-line patch allows me to do something like this:~

```
vim.g.toggle_quake_float = true
```

~in my `init.lua`. Then by setting~

```
float_opts = {
  height = math.floor(vim.o.lines * 0.4),
  width = vim.o.columns,
}
```

~I get a true Quake-style float terminal.~

~It didn't seem worthwhile to create an entirely new `direction` when it's such a tiny variation on the existing `float`, but there might be an argument to make it part of the `float_opts` table?~

~If you think this change is worth incorporating, I can add the corresponding paragraph to `toggleterm.txt`.~

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,9 +204,11 @@ require("toggleterm").setup{
     -- the 'curved' border is a custom border type
     -- not natively supported but implemented in this plugin.
     border = 'single' | 'double' | 'shadow' | 'curved' | ... other options supported by win open
-    -- like `size`, width and height can be a number or function which is passed the current terminal
+    -- like `size`, width, height, row, and col can be a number or function which is passed the current terminal
     width = <value>,
     height = <value>,
+    row = <value>,
+    col = <value>,
     winblend = 3,
     zindex = <value>,
     title_pos = 'left' | 'center' | 'right', position of the title of the floating window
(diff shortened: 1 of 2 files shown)


akinsho:
This plugin doesn't use `g` variables so that won't be a good way to go here. Also tbh I think this use case is quite specific and something a user could build into their config not something that needs to ship with this plugin

penryu:
I agreed: I did have my concerns about the use of `g` vars, and a key in `float_opts` would probably fit a lot better.

When you say "build it into their config", are you talking about a configuration of toggleterm, or a different solution entirely?

And my granted my specific use case is probably very specific. The name of the variable I used especially.

With that in mind, is there any interest in a more generalized solution like:

```lua
{
  ...
  float_opts = {
    horiz_align = 'left' | 'center' | 'right',  -- default: 'center'
    vert_align = 'top' | 'center' | 'bottom',  -- default: 'center'
  },
  ...
}
```

penryu:
Nevermind. I just realized that `row` and `col` are also valid `float_opts`, and looked at how `M._resolve_size()` works. So yes, this is truly trivial to implement with:

```lua
float_opts = { row = 0, ... },
```

Here is a doc-only change instead. :)

akinsho (approved):
LGTM thanks

PR #534, opened 2024-01-18
feat: enabling title for floating terminals

Blackmorse:
Got an idea from #504 PR

- title is enabled when `display_name` is provided
- will not fail on NeoVim versions prior to 0.9
- added documentation entries
- `title_pos` is under more appropriate  `float_opts` section


akinsho (approved):
LGTM 👍🏾

PR #537, opened 2024-02-01
fix: add rate limit for repeated ToggleTerm trigger

ysy:
Issue: when terminal_mapping is enabled and key-mapping is 'Ctrl + \', if I hold 'Ctrl + \' for long time, ToggleTerm command will be triggerd repeatly(show and hide for floating terminal). If the combo-key(Ctrl + \) is released slightly slower(depend on keyboard), ToggleTerm will be triggered twice, causing the floating terminal window to briefly appear and then immediately hide again.

To address the above issue, I have implemented a delay mechanism that checks the time difference between the current ToggleTerm and the previous one. If the difference is less than 150ms, it will be ignored.

This modification is quite dirty, and I hope it can serve as a starting point for further improvements.

### DIFF
--- a/lua/toggleterm.lua
+++ b/lua/toggleterm.lua
@@ -253,7 +253,16 @@ function M.send_lines_to_terminal(selection_type, trim_spaces, cmd_data)
   api.nvim_win_set_cursor(current_window, { start_line, start_col })
 end
 
+local last_toggle_sec = 0
+local last_toggle_nano_sec = 0
 function M.toggle_command(args, count)
+  local sec, nano_sec = vim.loop.gettimeofday()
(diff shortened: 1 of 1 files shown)


akinsho:
Hi @ysy, I'm not sure why this is necessary at all? why are holding down the terminal toggle mapping so much? I really can't envision the use case for this. I developed this plugin around the idea that a person would create one terminal then maybe a little while later make another and so on. I don't know why the key would need to be spammed?

I think this might be too specific a use case to get added to the plugin itself

akinsho:
This also seems like something you could just add to your vim config rather than here. I don't think many people share this issue.

PR #545, opened 2024-02-16
User configuration for winfixwidth and winfixheight

raffaem:


raffaem:
Thanks for looking at this.

Sorry @akinsho , but I need `<CTRL-w>=` to work.

Do you know how can I make it work without those contributions?

How toggleterm managers the shell size?

How can I change the size using toggleterm?

How can I have all the windows with equal width and height using toggleterm?

akinsho:
You can use a function to set the size and width of the windows. That isn't maybe as seamless as the mapping but I'd rather not go down that rabbit hole as it's can of worms and this is frankly the first time someone has requested this use case (I'm not much for it personally).

raffaem:
> You can use a function to set the size and width of the windows. That isn't maybe as seamless as the mapping but I'd rather not go down that rabbit hole as it's can of worms and this is frankly the first time someone has requested this use case (I'm not much for it personally).

But a function within the config?

That doesn't allow you to change the size after the terminal has been initialized, right?

PR #553, opened 2024-03-14
fix: autochdir for custom terminals

ttytm:
Currently, autochdir won't work for custom terminals.

So currently I encounter scenarios like the following: Opening a custom terminal with i.e lazygit in a git project then closing the terminal, switching buffers to another git project and re toggling it, keeps opening lazygit in the first directory.

The changes in the PR aim to make this functionality work - switching buffers in the same session to another git project and opening a custom terminal should open it with the new git projects working directory.


diegodario88:
+1 waiting for this fix

ttytm:
Ping @akinsho. Might be a trivial one.

akinsho (approved):
LGTM 👍🏾

PR #555, opened 2024-03-21
feat: add the command ToggleTermBufferDir to create and/or toggle a terminal with the current buffer directory

TimaSv-Uk:
This PR introduces the :ToggleTermBufferDir command, which allows users to switch the buffer directory in the terminal. If a terminal instance with a buffer directory wasn’t already there, one is created. However, once a terminal instance is created, the buffer terminal name or direction cannot be changed.

The :TermSelect command can be used to utilize toggleterm from another buffer in the current buffer. By default, the **name** is set to PATH_TO_CURRENT_BUFFER and the **direction** is set to float. The size is taken from the setup function.

Here is an example usage of the command:
`
:ToggleTermBufferDir name=desktop direction=horizontal`

I love to use Harpoon or Telescop, and the new command allows me to quickly open Terminla after jumping to a new project. If there are any issues, I'll try to fix them.

Uploading ToggleTermBufferDir_example.mp4…

### DIFF
--- a/README.md
+++ b/README.md
@@ -253,7 +253,7 @@ when using `TermSelect` command to indicate the specific terminal.
 `size` and `direction` are ignored if the command closes a terminal.
 
 
-### `TOGGLETERMTOBUFFERDIR`
+### `TOGGLETERMBUFFERDIR`
 
 This command switches the buffer directory in the terminal.
 If a terminal instance with a buffer directory wasn't already there,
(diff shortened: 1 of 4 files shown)


akinsho:
@TimaSv-Uk thanks for the PR. In general I strongly try to avoid increasing the surface area of my plugins by adding more and more commands. In this case (apologies as I only skimmed this PR) I don't understand exactly why this is needed. A user can just run `cd MY_DIR` once the terminal is open. Why does this need to be a builtin command?

rafisics:
> As a university student, you’re often working on various small programs in different languages, like Python, C#, and others. It can indeed be a pain to have to navigate into the directory every time you want to run a console application, especially when you need to show and run several scripts one after another to your professor. 

Have you found any solution?

PR #557, opened 2024-03-30
Add string array support to `open_mapping` setting.

sbkkalex:
I sometime use a Japanese keyboard, and it don't have the `\` key, instead it have a `¥` key.

It would be helpful for people like me who what to bind multiple keys for toggle terminal. So I modified the `open_mapping` setting to also take an array of string as parameter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,7 +126,7 @@ I'm also going to be pretty conservative about what I add.
 
 This plugin must be explicitly enabled by using `require("toggleterm").setup{}`
 
-Setting the `open_mapping` key to use for toggling the terminal(s) will set up mappings for _normal_ mode.
+Setting the `open_mapping` key to use for toggling the terminal(s) will set up mappings for _normal_ mode. The `open_mappint` can be a key string or an array of key strings. 
 If you prefix the mapping with a number that particular terminal will be opened. Otherwise if a prefix is not set, then the last toggled terminal will be opened. In case there are multiple terminals opened they'll all be closed, and on the next mapping key they'll be restored.
 
 If you set the `insert_mappings` key to `true`, the mapping will also take effect in insert mode; similarly setting `terminal_mappings` to `true` will have the mappings take effect in the opened terminal.
@@ -161,7 +161,7 @@ require("toggleterm").setup{
       return vim.o.columns * 0.4
     end
   end,
-  open_mapping = [[<c-\>]],
+  open_mapping = [[<c-\>]], -- or { [[<c-\>]], [[<c-¥>]] } if you also use a Japanese keyboard.
   on_create = fun(t: Terminal), -- function to run when the terminal is first created
   on_open = fun(t: Terminal), -- function to run when the terminal opens
   on_close = fun(t: Terminal), -- function to run when the terminal closes
--- a/README.md
+++ b/README.md
@@ -126,7 +126,7 @@ I'm also going to be pretty conservative about what I add.
 
 This plugin must be explicitly enabled by using `require("toggleterm").setup{}`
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -126,7 +126,7 @@ I'm also going to be pretty conservative about what I add.
 
 This plugin must be explicitly enabled by using `require("toggleterm").setup{}`
 
-Setting the `open_mapping` key to use for toggling the terminal(s) will set up mappings for _normal_ mode.
+Setting the `open_mapping` key to use for toggling the terminal(s) will set up mappings for _normal_ mode. The `open_mappint` can be a key string or an array of key strings.
```
akinsho, line 129:
```suggestion
Setting the `open_mapping` key to use for toggling the terminal(s) will set up mappings for _normal_ mode. The `open_mapping` can be a key string or an array of key strings. 
```

doc/toggleterm.txt
```diff
@@ -102,11 +102,11 @@ SETUP ~
 This plugin must be explicitly enabled by using `require("toggleterm").setup{}`
 
 Setting the `open_mapping` key to use for toggling the terminal(s) will set up
-mappings for _normal_ mode. If you prefix the mapping with a number that
-particular terminal will be opened. Otherwise if a prefix is not set, then the
-last toggled terminal will be opened. In case there are multiple terminals
-opened they’ll all be closed, and on the next mapping key they’ll be
-restored.
+mappings for _normal_ mode. The `open_mappint` can be a key string or an array
```
akinsho, line 105:
```suggestion
mappings for _normal_ mode. The `open_mapping` can be a key string or an array
```

lua/toggleterm.lua
```diff
@@ -35,19 +35,29 @@ end
 
 local function setup_global_mappings()
   local mapping = config.open_mapping
-  -- v:count defaults the count to 0 but if a count is passed in uses that instead
-  if mapping then
-    vim.keymap.set("n", mapping, '<Cmd>execute v:count . "ToggleTerm"<CR>', {
+
+  -- Key mapping function
+  ---@param key string
+  local map = function(key)
```
akinsho, line 41: I prefer the style

```suggestion
  local function map(key)
```

lua/toggleterm.lua
```diff
@@ -35,19 +35,29 @@ end
...
     if config.insert_mappings then
-      vim.keymap.set("i", mapping, "<Esc><Cmd>ToggleTerm<CR>", {
+      vim.keymap.set("i", key, "<Esc><Cmd>ToggleTerm<CR>", {
         desc = "Toggle Terminal",
         silent = true,
       })
     end
   end
+
+  if type(mapping) == "string" then map(mapping) end
```
akinsho, line 55: Although it's impossible (🤞🏾) I'd rather this was an `if then else if` I feel uncomfortable with the idea that it could possible fall into the next clause even though it can't

lua/toggleterm/terminal.lua
```diff
@@ -132,8 +132,20 @@ end
...
+    vim.keymap.set("t", key, "<Cmd>ToggleTerm<CR>", { buffer = bufnr, silent = true })
+  end
+
+  if config.terminal_mappings then
+    if type(mapping) == "string" then map(mapping) end
+    if type(mapping) == "table" then
+      for _, key in pairs(mapping) do
+        map(key)
+      end
+    end
```
akinsho, line 148: Looks to me like this logic is just being repeated maybe it's best to do this once in `utils.lua` generically and use that helper function in both of the places you've made this change

akinsho (changes requested):
@sbkkalex thanks for this PR, happy to have this more flexible to accommodate other folks 👍🏾, I've just made a few comments mainly around typos and the duplication

sbkkalex:
@akinsho 
Thank you for your advise. I extracted the logic to utils.

akinsho (approved):

PR #561, opened 2024-04-07
Correct CR for Nushell

YuanYuYuan:
As [reedline stating in their doc](https://github.com/nushell/reedline/tree/d2e0d320a9258b632c4470cc9d7637d1f132916a/src#technical-background), nushell uses the same CRLF policy like Windows. This PR adds the support for it.


akinsho (approved):
LGTM thanks

PR #566, opened 2024-04-12
fix: ensure `on_choice` operates on exact `items` element

echasnovski:
As `on_choice()` actions depend on modifying terminal objects in place, make sure to operate on exact element and not a copy. As users can override `vim.ui.select()`, its implementation might return an item's copy (which is not explicitly prohibited) making later `on_choice()` actions not 100% full proof.


akinsho (approved):
Thanks 👍🏾

Issue #567, opened 2024-04-14
[Feature Request] Support Vim Sessions


ColinKennedy:
If you use Vim Sessions, e.g. `:mksession Session.vim` and then `nvim -S Session.vim`, the terminal buffer is restored but toggleterm.nvim doesn't recognize the restored terminal buffer as a "toggleterm" buffer. So the next `:ToggleTerm` call creates a second terminal. There's also other issues as well, for example the restored terminal won't have tte same styling as if it were created using `:ToggleTerm`.

Being able to restore the toggleterm terminal will likely require https://github.com/akinsho/toggleterm.nvim/issues/528 so I'd consider the two issues to be related.

chancez:
I was going to create an issue for this as well. I think the main issue is all the state is stored directly in the toggleterm plugin. I think for this to work as expected, toggleterm needs to associate it's state with the buffer for a terminal. This would mean storing the resulting data it maintains on the `vim.b` variable for a terminal buffer.

PR #568, opened 2024-04-14
Add session support

ColinKennedy:
Closes: https://github.com/akinsho/toggleterm.nvim/issues/567

Note: Currently we assume that we are the only owners of a `Sessionx.vim` file. Since (Neo)vim only allows one of these files at a time, we have to clear the contents completely. In the future ideally multiple x.vim files will be allowed. See https://github.com/akinsho/toggleterm.nvim/issues/567

### DIFF
--- a/lua/toggleterm/terminal.lua
+++ b/lua/toggleterm/terminal.lua
@@ -184,6 +184,36 @@ local function _get_dir(dir)
   end
 end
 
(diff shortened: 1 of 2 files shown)


ColinKennedy:
This PR was a bit premature. It doesn't work with 2+ terminals in the same tab. Will need to fix and reopen.

PR #572, opened 2024-04-21
Fix column indexing

raffaem:
We started in 75d3de9d261431dd4d6a68134bb46907c91c2023, this should finish the job by restoring correct behavior for "single_line" motion type.


akinsho (approved):
LGTM (from a distance)

Issue #597, opened 2024-08-04
[question] Is there a way to make clear_env take effect by default


fcying:
Is there a way to make clear_env take effect by default? I tried configuring `t.clear_env = true` in on_create, but it didn't work for run `ToggleTerm`

```
on_create = function(t)
    t.clear_env = true
end,
```

PR #598, opened 2024-08-05
feat: add clear_env option

fcying:
add clear_env option for #597


akinsho:
Seems reasonable thanks for the PR

Issue #616, opened 2024-10-18
Is there a way to stack terminals on top of each other ?


searleser97:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

when opening multiple terminals, they all appear at the bottom next to each other, when having multiple of them opened at the same time, is very hard to see the actual contents due to the narrow space left in each.

### Expected Behavior

I would like to have them opened at full maximum width one on top of each other so that I can see the contents clearly and at the same time.

### Steps To Reproduce

1. Toggle multiple terminals with default configuration

### Environment

```Markdown
- OS:
- neovim version:
- Shell:
```

searleser97:
Friendly ping

akinsho:
@searleser97 the formats mentioned in the README and shown in the docs are all that are supported e.g. horizontal, vertical etc. I'm not currently looking to add more layouts tbh although a well thought through proposal for more that you'd be willing to implement is something I'd be happy to review

searleser97:
@akinsho here is the proposal https://github.com/akinsho/toggleterm.nvim/pull/618

PR #618, opened 2024-10-30
Add Responsive Capabilities for horizontal direction

searleser97:
# Context
When using Neovim with toggleterm in a vertical monitor there is not enough space to see the output of 2 or more terminals that are showed at the same time properly due to the amount of space left in it. Like in the following image

</img>

It would be better to have them stacked on top of each other so that the user can see clearly the output of each of them when using vertical monitor, or just a window with smaller width. Like in the following image

</img>

#closes https://github.com/akinsho/toggleterm.nvim/issues/616

# What does this PR do?

This PR solves the problem by adding a configuration setting that can be used to be define at which point the user would like to start stacking terminals on top of each other, instead of creating them next to each other when wanting to visualize more than one at the same time. The behavior described is better illustrated in the following video.

https://github.com/user-attachments/assets/4aaa5b05-17a0-44bd-ae4c-77ef6fd094e9

### DIFF
--- a/lua/toggleterm/config.lua
+++ b/lua/toggleterm/config.lua
@@ -14,6 +14,9 @@ local function shade(color, factor) return colors.shade_color(color, factor) end
 ---@field name_formatter fun(term: Terminal):string
 ---@field enabled boolean
 
+--- @class ResponsivenessSettings
+--- @field horizontal_breakpoint number
+
 --- @class ToggleTermConfig
 --- @field size number
 --- @field shade_filetypes string[]
@@ -37,6 +40,7 @@ local function shade(color, factor) return colors.shade_color(color, factor) end
 --- @field winbar WinbarOpts
 --- @field autochdir boolean
 --- @field title_pos '"left"' | '"center"' | '"right"'
+--- @field responsiveness_settings ResponsivenessSettings
 
 ---@type ToggleTermConfig
 local config = {
@@ -65,6 +69,9 @@ local config = {
     winblend = 0,
     title_pos = "left",
   },
+  responsiveness_settings = {
+    horizontal_breakpoint = 135,
+  }
 }
 
 ---Derive the highlights for a toggleterm and merge these with the user's preferences
--- a/lua/toggleterm/config.lua
+++ b/lua/toggleterm/config.lua
@@ -14,7 +14,7 @@ local function shade(color, factor) return colors.shade_color(color, factor) end
 ---@field name_formatter fun(term: Terminal):string
 ---@field enabled boolean
 
---- @class ResponsivenessSettings
+--- @class Responsiveness
 --- @field horizontal_breakpoint number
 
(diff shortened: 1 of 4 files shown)


lua/toggleterm/ui.lua
```diff
@@ -316,7 +317,14 @@ function M.open_split(size, term)
     local split_win = windows[#windows]
     if config.persist_size then M.save_window_size(term.direction, split_win.window) end
     api.nvim_set_current_win(split_win.window)
-    vim.cmd(commands.existing)
+    local window_width = vim.o.columns
+    local horizontal_breakpoint =
+      require("toggleterm.config").get("responsiveness_settings").horizontal_breakpoint
```
akinsho, line 322: the config module is already imported above you only need to do `config.responsiveness_settings`

lua/toggleterm/config.lua
```diff
@@ -65,6 +69,9 @@ local config = {
     winblend = 0,
     title_pos = "left",
   },
+  responsiveness_settings = {
```
akinsho, line 72: I think the addition of the term `_settings` is verbose, it's already in a config object so I don't thing that's needed. Trying to think of how things like css or desktop apps refer to their layout/responsive features, `responsiveness` on it's own is ok in lieu of something better

akinsho (commented):
@searleser97 this looks great thanks for taking the time to contribute. Just some comments on naming, another thought I have is whether or not someone is going to want to turn this off because there's always someone who wants something weird like that. Rather than necessarily add a new option maybe add the ability to set `horizontal_breakpoint = false` so they can stop it from stacking

searleser97:
> @searleser97 this looks great thanks for taking the time to contribute. Just some comments on naming, another thought I have is whether or not someone is going to want to turn this off because there's always someone who wants something weird like that. Rather than necessarily add a new option maybe add the ability to set `horizontal_breakpoint = false` so they can stop it from stacking

@akinsho about this comment, user can turn this off by setting the breakpoint to 0

lua/toggleterm/ui.lua
```diff
@@ -318,8 +318,7 @@ function M.open_split(size, term)
     if config.persist_size then M.save_window_size(term.direction, split_win.window) end
     api.nvim_set_current_win(split_win.window)
     local window_width = vim.o.columns
-    local horizontal_breakpoint =
-      require("toggleterm.config").get("responsiveness_settings").horizontal_breakpoint
+    local horizontal_breakpoint = config.get("responsiveness").horizontal_breakpoint
```
akinsho, line 321: The `.get(` is also unnecessary you can just do `config.responsiveness.horizontal_breakpoint`

searleser97: Done

akinsho:
@searleser97 merged, thanks for the contribution ❤️

PR #622, opened 2024-12-07
fix: colorscheme on hidden terminal

### LINKED ISSUE #435: Error message when running the `colorscheme` command

### Current Behavior

Hi, this "bug" seems to be due to how I'm using the toggleterm API, but I'm not sure what I'm doing wrongly so any help would be much appreciated!

I'm using the `lazy.nvim` package manager, and in the `init` function for `toggleterm`, i have the following code:
```lua
M.init = function()
  local toggleterm_spawn = vim.api.nvim_create_augroup("toggleterm_spawn", {})
  vim.api.nvim_create_autocmd("VimEnter", {
    group = toggleterm_spawn,
    pattern = "*",
    callback = function()
      local Terminal = require("toggleterm.terminal").Terminal
      local term = Terminal:new()
      term:spawn()
    end,
  })
end
```
I'm doing this because for some reason WSL takes a long time to load up a new shell instance, so I wanted to "preload" a shell instance so that the shell loads up instantly when opening `toggleterm`.

However when I run `colorscheme [COLOR_SCHEME]`, I get the error:
```
Error detected while processing ColorScheme Autocommands for "*":
Error executing lua callback: ...l/share/nvim/lazy/nvim-toggleterm.lua/lua/toggleterm.lua:322: Expected Lua number
stack traceback:
	[C]: in function 'nvim_win_is_valid'
	...l/share/nvim/lazy/nvim-toggleterm.lua/lua/toggleterm.lua:322: in function <...l/share/nvim/lazy/nvim-toggleterm.lua/lua/toggleterm.lua:319>
```
I can avoid the error by changing
```lua
if api.nvim_win_is_valid(term.window) then
    api.nvim_win_call(term.window, function() ui.hl_term(term) end)
end
```
to
```lua
if term.window and api.nvim_win_is_valid(term.window) then
    api.nvim_win_call(term.window, function() ui.hl_term(term) end)
end
```
in `toggleterm.lua`.

Removing the `init` function shown above also removes the error. The colorscheme changes properly in spite of the error message shown.

I'm not sure whether this is an actual bug with the plugin, or a mistake on my part. If there is a best practice for achieving the "terminal preloading" functionality, I would love to know about it!

### Expected Behavior

There is no error when running the `colorscheme` command.

### Steps To Reproduce

1. Install `toggleterm` through `lazy.nvim` with the following spec:
```lua
local M = {
  "akinsho/nvim-toggleterm.lua",
  keys = { [[<C-\>]] },
}

M.init = function()
  local toggleterm_spawn = vim.api.nvim_create_augroup("toggleterm_spawn", {})
  vim.api.nvim_create_autocmd("VimEnter", {
    group = toggleterm_spawn,
    pattern = "*",
    callback = function()
      local Terminal = require("toggleterm.terminal").Terminal
      local term = Terminal:new()
      term:spawn()
    end,
  })
end

M.config = function()
  local toggleterm_height = math.ceil(vim.api.nvim_win_get_height(0) * 0.45)

  require("toggleterm").setup({
    size = toggleterm_height,
    open_mapping = [[<C-\>]],
    shade_terminals = true,
    shading_factor = "1",
    start_in_insert = true,
    persist_size = true,
    direction = "horizontal",
    autochdir = true,
  })
end

return M
```
2. Run `colorscheme ...` for any colorscheme

### Environment

```Markdown
- OS: macos 13.3
- neovim version:
NVIM v0.10.0-dev-36+g7ffe45017
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3

- Shell: fish shell 3.6.1
```

### Anything else?

_No response_

Foo-x:
Fixes #435

I spawn a terminal at startup to improve initial performance. But if I run `colorscheme` before the terminal is shown, i.e. before it's attached to a window, an error is thrown because `term.window` is nil. I have fixed to verify `term.window` is not nil.

### DIFF
--- a/lua/toggleterm.lua
+++ b/lua/toggleterm.lua
@@ -351,7 +351,7 @@ local function setup_autocommands(_)
     callback = function()
       config.reset_highlights()
       for _, term in pairs(terms.get_all()) do
-        if api.nvim_win_is_valid(term.window) then
+        if term.window and api.nvim_win_is_valid(term.window) then
           api.nvim_win_call(term.window, function() ui.hl_term(term) end)
         end
       end


GustavEikaas:
Thank you for fixing this @Foo-x. I needed the exact same thing!

PR #623, opened 2024-12-24
feat: Add TermNew command

waiting-for-dev:
Humans are bad at remembering numbers, so this command allows you to open a new terminal at the next available count. It's helpful in combination with `:TermSelect` to work with terminals without needing to remember numbers. It supports the regular arguments as `:ToggleTerm`.

### DIFF
--- a/lua/toggleterm/terminal.lua
+++ b/lua/toggleterm/terminal.lua
@@ -106,7 +106,7 @@ local Terminal = {}
 --- hasn't already been allocated e.g. in a list of {1,2,5,6} the next id should
 --- be 3 then 4 then 7
 ---@return integer
-local function next_id()
+function M.next_id()
   local all = M.get_all(true)
   for index, term in pairs(all) do
     if index ~= term.id then return index end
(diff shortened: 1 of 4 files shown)


akinsho (approved):
Sorry for the long delay this seems reasonable to me, haven't looked at the code in a while so hopefully there isn't some assumption this breaks

akinsho:
Think the testing workflow need some maintenance, probably not the code breaking

lua/toggleterm/terminal.lua
```diff
@@ -586,7 +586,7 @@ if _G.IS_TEST then
     end
   end
 
-  M.__next_id = next_id
```
akinsho, line 589: I actually think this isn't necessary anymore. Since I wasn't exporting `next_id` before this was a way of trying to do it privately so I could use it in tests but now it's public this line can be removed and the corresponding usages of `M.__next_id` can just be `M.next_id`

W3ND31:
Hello, @akinsho ! You think in create a new release/tag for this option? The last one is from December (as the last commit behind this)

PR #631, opened 2025-02-24
Fix persiste-mode when leaving term by gf

### LINKED ISSUE #630: [BUG] persiste-mode not working when leaving term by gf movement
`persist-mode` is handled on `WinLeave` event, so it does not work when leaving term by `gf` movement not closing the window.

zhengpd:
close #630

### DIFF
--- a/lua/toggleterm.lua
+++ b/lua/toggleterm.lua
@@ -334,7 +334,7 @@ local function setup_autocommands(_)
     callback = handle_term_enter,
   })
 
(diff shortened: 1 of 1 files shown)


akinsho:
@zhengpd have you tested to see if this breaks other workflows, this seems like the sort of change that works for one use case but breaks a bunch of others?

zhengpd:
@akinsho I didn't see any problem with my daily uses, however I don't have time for more deep testing recently. I'll close this for now.

